PR #1, opened 2021-05-20
OPTM-27: task list widget

firstdayofjune:
- removed floating action button from home screen (since it was also visible in all other screens)
- added large plus button to task_list_screen to add tasks
- tasks can now be checked/unchecked
- task_form_screen now has an "edit" mode, to distinguish editing a task from creating one
- added a placeholder text when no tasks are there (for better usability)
- refactorings
- added ios & files


.github/workflows/main.yaml
```diff
@@ -9,6 +9,7 @@ jobs:
       - uses: subosito/flutter-action@v1
         with:
           flutter-version: '2.0.6'
+      - run: flutter clean
```
JacobRoe: You almost never need a flutter clean run in the workflow because the jobs always start with a fresh environment. The command `flutter clean` deletes the build cache, which isn't present on a fresh environment.

lib/screens/task_list_screen.dart
```diff
@@ -59,4 +83,24 @@ class _TaskListScreenState extends State<TaskListScreen> {
...
     );
   }
+
+  Widget _buildTaskContainer(Task task, int index) {
+    return Container(
+      color: index.isEven ? Colors.white : Colors.white30,
+      child: ListTile(
+        leading: task.completed
+            ? Icon(Icons.check_circle, color: Colors.green)
+            : Icon(Icons.circle_outlined),
```
JacobRoe: Circle outlined is indeed not available in the current channel. (Although it's listed in the flutter docs.)

See also the issues:
* https://github.com/flutter/flutter/issues/27956, 
* https://github.com/flutter/flutter/pull/61778
* https://github.com/flutter/flutter/issues/82349
* etc.

Two possible solutions would be:
* using the package material_design_icons_flutter
* switching to a different channel

Regarding the latter option they recommend to use the stable channel (which we are in currently). That said, they also state using the beta channel should be fine because it serves just as an additional maturing phase for a stable release without any extra testing involved. See https://github.com/flutter/flutter/wiki/Flutter-build-release-channels for more.

I just tried it out with the beta channel and it seems to work. The question remains, should we switch into the beta channel or better use the package? (Theoretically we could also just include the material icon font ourselves as an asset and used it from there.) What's your opinion on that matter?

JacobRoe (changes requested):
Looks amazing! Especially the refactoring is very much appreciated.

Still, we need to decide how to proceed regarding the icon problem and the clean command in the workflow should be removed.

JacobRoe (approved):

.github/workflows/main.yaml
firstdayofjune: Ok, that's what I thought, but I wasn't sure, so I gave it a try :-) Thanks for confirming though, I removed the clean statements.

lib/screens/task_list_screen.dart
firstdayofjune: Thanks for your input. For the moment, I changed the Icon to a `radio_button_off` icon and added a semanticLabel to make the meaning of the icon also clear for screen readers. I also have the rational for this change in the commit message, so git blame will explain the change. I just realized, I could have added a comment in the code as well, which I am happy to add with my next commit.

PR #3, opened 2021-05-24
OPTM-36 Removed pubspec.lock from git

guisedlythe:


firstdayofjune:
The dart documentation recommends to keep pubspeck.lock under version control for application packages: https://dart.dev/guides/libraries/private-files#pubspeclock, so that everyone uses the same version of the packages. This also ensures that the CI tests run against the same package versions as the local build. Further reasoning for keeping the pubspeck.lock is also discussed here: https://github.com/dart-lang/site-www/issues/2408

I therefore vote for keeping the file under version control. If a current dependency breaks the build (on Windows), commit the updated "working" lock-file and let's find out, if this works on all other OSes as well (also, the diff might tell us which package caused an issue on Windows).

PR #6, opened 2021-06-02
OPTM-48 add description

Tan-jay:


lib/screens/task_form_screen.dart
```diff
@@ -16,12 +16,16 @@ class _TaskFormScreenState extends State<TaskFormScreen> {
...
+  late final _formDescriptionController =
+      TextEditingController(text: _task.description);
   bool _editing = false;
 
   void _onSubmitPressed(context) {
     if (_formKey.currentState!.validate()) {
       Provider.of<TaskListService>(context, listen: false)
           .upsert(_task..title = _formController.text);
+      Provider.of<TaskListService>(context, listen: false)
+          .upsert(_task..description = _formDescriptionController.text);
```
JacobRoe: You could reuse the previous provided service + use the cascade operator (..) of Dart like e.g.:
```dart
final taskListService = Provider.of<TaskListService>(context, listen: false);
taskListService.upsert(_task
    ..title = _formController.text
    ..description = _formDescriptionController);
```

* https://dart.dev/guides/language/language-tour#cascade-notation

lib/screens/task_list_screen.dart
```diff
@@ -102,7 +106,7 @@ class _TaskListScreenState extends State<TaskListScreen> {
               ),
         title: Text(task.title),
         // ToDo Change subtitle to task description
-        subtitle: Text(task.title),
+        subtitle: Text(description),
```
JacobRoe: You could use the syntactic sugar of Dart like e.g.
```dart
subtitle: Text(task.description ?? ''),
```

* https://dart.dev/guides/language/language-tour#conditional-expressions

JacobRoe (changes requested):
Nice, very good!

You might want to address my comments. Mandatory for an approval would be the part about reusing the previous provider in _onSubmitPressed. Otherwise, it's on point.

Tan-jay:
> Nice, very good!
> 
> You might want to address my comments. Mandatory for an approval would be the part about reusing the previous provider in _onSubmitPressed. Otherwise, it's on point.

done, thanks! :)

JacobRoe:
> > Nice, very good!
> > You might want to address my comments. Mandatory for an approval would be the part about reusing the previous provider in _onSubmitPressed. Otherwise, it's on point.
> 
> done, thanks! :)

Looks very good, thank you!

PR #7, opened 2021-06-03
OPTM-23 activity log UI

firstdayofjune:
Discussing @bithianjuman changes to the UI.


lib/screens/activity_log_screen.dart
```diff
@@ -37,6 +39,7 @@ class _ActivityLogScreenState extends State<ActivityLogScreen> {
       ));
     }
     return ListView.separated(
+      padding: EdgeInsets.symmetric(vertical: 8),
```
firstdayofjune: Is this to give the Entries some more spacing?

lib/screens/activity_log_screen.dart
```diff
@@ -46,11 +49,94 @@ class _ActivityLogScreenState extends State<ActivityLogScreen> {
...
+        fontWeight: FontWeight.bold,
+        color: Colors.black,
+        fontSize: 18,
+      ),
+    );
+  }
+
+  Widget _getActivityDate(ActivityEntry activity) {
+    return Text(
+      "${activity.timestamp?.day}/${activity.timestamp?.month}/${activity.timestamp?.year}",
```
firstdayofjune: This formats the date disregarding of the apps localization. I don't mind this for the moment, but could you check if there is something like a localized print method for dates or an existing package which prints dates in various formats depending on the localization? So e.g. a package which prints the date in YYYY/MM/D if language is EN and D.MM.YYYY if language is DE?

lib/screens/activity_log_screen.dart
```diff
@@ -46,11 +49,94 @@ class _ActivityLogScreenState extends State<ActivityLogScreen> {
...
+      style: TextStyle(
+        color: Colors.black.withOpacity(0.6),
+        fontSize: 14,
+      ),
+    );
+  }
+
+  Widget _getActivityBasedIcon(ActivityEntry activity) {
```
firstdayofjune: I would call it `getActivityIcon`, it is clear that the icon is based on the activity bc. we pass the ActivityEntry as an argument :-)

lib/screens/activity_log_screen.dart
```diff
@@ -46,11 +49,94 @@ class _ActivityLogScreenState extends State<ActivityLogScreen> {
...
+    return Image.asset(
+      activity.event == "Task Completed"
```
firstdayofjune: The Icon should be toggled depending on the device, where the activity comes from. If the event was created on this device, it is an outgoing icon, if it comes from any other device, show the download icon.

lib/screens/activity_log_screen.dart
```diff
@@ -46,11 +49,94 @@ class _ActivityLogScreenState extends State<ActivityLogScreen> {
...
+          ? "./assets/up_arrow_square.png"
```
firstdayofjune: Please use matching material icons from https://fonts.google.com/icons instead of adding .PNGs (e.g. `Icon(Icons.file_download_outlined)`)

lib/screens/activity_log_screen.dart
```diff
@@ -46,11 +49,94 @@ class _ActivityLogScreenState extends State<ActivityLogScreen> {
...
+          : "./assets/download.png",
```
firstdayofjune: Please use matching material icons from https://fonts.google.com/icons instead of adding .PNGs.

lib/screens/activity_log_screen.dart
```diff
@@ -46,11 +49,94 @@ class _ActivityLogScreenState extends State<ActivityLogScreen> {
...
+      color: Colors.black,
+      height: 22,
+      width: 22,
+    );
+  }
+
+  Widget _getActivityDesc(ActivityEntry activity) {
```
firstdayofjune: Would make it more verbose and name it: `_getActivityDescription`

lib/screens/activity_log_screen.dart
```diff
@@ -46,11 +49,94 @@ class _ActivityLogScreenState extends State<ActivityLogScreen> {
...
+    return RichText(
+      text: TextSpan(
+        text: "On ",
```
firstdayofjune: The description text should indicate which device initiated the change. There are two scenarios, where activities can come from. Either they are created on "this device", (A) so the currently used device, or they were synced from a different device (B).

In case of A, show: On this device in **\<Name of the TaskList>**.
In case of B, show: On **\<device name>** in **\<Name of the TaskList>**.

lib/screens/activity_log_screen.dart
```diff
@@ -46,11 +49,94 @@ class _ActivityLogScreenState extends State<ActivityLogScreen> {
...
+                  ? FontWeight.normal
+                  : FontWeight.bold,
+            ),
+          ),
+          TextSpan(
+            text: " in ",
+            style: TextStyle(color: Colors.black.withOpacity(0.5)),
+          ),
+          TextSpan(
+            text: activity.network,
```
firstdayofjune: Instead of network, use the device here.

lib/screens/task_list_screen.dart
```diff
@@ -95,7 +95,7 @@ class _TaskListScreenState extends State<TaskListScreen> {
                 semanticLabel: "Completed Task",
               )
             : Icon(
-                Icons.circle_outlined,
```
firstdayofjune: Why the change from circle_outlined to circle?

pubspec.lock
```diff
@@ -28,7 +28,7 @@ packages:
       name: async
       url: "https://pub.dartlang.org"
     source: hosted
-    version: "2.6.1"
+    version: "2.5.0"
```
firstdayofjune: I see some packages are downgraded again, did you run `flutter pub get` before your commit?

pubspec.yaml
```diff
@@ -57,9 +57,9 @@ flutter:
   uses-material-design: true
 
   # To add assets to your application, add an assets section, like this:
-  # assets:
-  #   - images/a_dot_burr.jpeg
-  #   - images/a_dot_ham.jpeg
+  assets:
```
firstdayofjune: As mentioned above, please remove the images again in favor of the flutter material icons.

lib/screens/activity_log_screen.dart
bithianjuman: yes

lib/screens/activity_log_screen.dart
bithianjuman: i am not fully aware of the localization.

lib/screens/activity_log_screen.dart
bithianjuman: can you please explain me again .I didn't get your point?

lib/screens/activity_log_screen.dart
firstdayofjune: I also don't know, how localization works in Flutter. Judged by the maturity of flutter I am confident some ready to use solutions exists, though. Could you please google for some best practices? Maybe this even is a standard feature of flutter? I don't think it should be too complicated to do this. Feel free to add any 3rd party packages, if this makes it easier.

lib/screens/activity_log_screen.dart
firstdayofjune: We build a task list app. A user can create multiple task lists (currently we have only one). Each list has a name and contains tasks. Whenever a task is created or changed an activity should be logged and displayed. Task lists can be synced via P2P. When two devices sync, they exchange the tasks they have in the list. This may cause changes to tasks (e.g. "Task created", "Description Updated", "Task completed"...). These changes should be logged in activites.

The title of an activity describes what was changed (e.g. "Task created", "Description Updated", "Task completed"...).
The description of an activity should give additional context info:

- In which list did the update occur? (Name of the TaskList)
- In case of a sync, which device does the change come from (\<device name> or "this device")

The displayed activity description therefore changes depending on whether:

- The activity relates to a task that was changed on the currently used device (Scenario A)
- The activity relates to a task synced from a different device (Scenario B)

lib/screens/activity_log_screen.dart
bithianjuman: I try to find something.

lib/screens/activity_log_screen.dart
bithianjuman: done

lib/screens/activity_log_screen.dart
bithianjuman: done

lib/screens/activity_log_screen.dart
bithianjuman: done

pubspec.lock
bithianjuman: I think the versions are not matter, it should work fine if we stay with this version.

pubspec.yaml
bithianjuman: done

lib/screens/task_list_screen.dart
bithianjuman: Actually current version giving me error so I have used another library from flutter icon.

lib/screens/activity_log_screen.dart
bithianjuman: done

pubspec.lock
firstdayofjune: If there is no specific reason to downgrade a package, we should not do that. By downgrading a version we may even loose functionality in external libraries and by such break functionality without noticing. Please revert your changes to the pubspec.lock and run `flutter pub get` to install the latest packages.

lib/screens/task_list_screen.dart
firstdayofjune: What was the error you had? Some of the Icons are only supported with the latest flutter version, so try `flutter upgrade` on your device and check if the default flutter icons work. I would prefer using the built in icons, instead of adding another dependency on Eva Icons, here.

guisedlythe:
I just quickly looked into this PR and it looks good for important parts.

Some minor things:
- There are some minor code formatting issues with our new code analysis. You can run `flutter pub run dart_code_metrics:metrics lib` and fix the warnings in your code (I am only referring to the "Missing blank line before return" and "Prefer trailing comma" rule, there are other warnings in the code that is already present which we may fix later)
- In my opinion there should be a blank line after the member variables (in `activity_log_screen.dart:16` and `activity_entry.dart:19`)
- Remove the unused assets `up_arrow_square.png` and `download.png` (assuming they won't be used in the very near future)
- Instead of comparing to "Windows Phone" to check if this is your own device, you could add a method to ActivityEntryService called something like getCurrentDeviceName() that just returns "Windows Phone" for now. You can than use this method everywhere in the UI activity log screen to check for your own device.
- Since the activity log screen doesn't need any state (and I assume it won't need any in the future), you can just make it into a StatelessWidget (c.f. TaskListScreen)
- About the changes in `pubspec.lock`:
  - When I run `flutter pub get` some version numbers in  `pubspec.lock` change slightly
  - Might be related to the flutter version. What flutter version do you use (I am using 2.2.1 with dart 2.13.1)?

Related things that should not affect this PR but we might want to discuss:
- We might want to store the device ID in ActivityEntry.device and use the PeerInfoService to get the corresponding device name
- Because we query pretty much all our state from the database, which always yields a Future<...> object, we might need to wrap the activity entry list in a FutureBuilder to wait for the state to be ready. This is a design decision that originated from @JacobRoe  s implementation of the Database and we just quietly decided to stick to it. There are benefits and drawbacks to this approach, but I would suggest that we just stick to it.

lib/screens/activity_log_screen.dart
```diff
@@ -0,0 +1,154 @@
...
+        fontWeight: FontWeight.normal,
+        color: Colors.black,
+        fontSize: 18,
+      ),
+    );
+  }
+
+  Widget _getActivityDate(ActivityEntry activity) {
+    return Text(
+      '${activity.timestamp?.day}/${activity.timestamp?.month}/${activity.timestamp?.year}',
```
Tan-jay: you could also use a formatting package :)  like 
` DateFormat('dd.MM.yyyy').format(activity.timestamp!),`
from 'package:intl/intl.dart'

lib/screens/activity_log_screen.dart
```diff
@@ -0,0 +1,154 @@
...
+      },
+    );
+  }
+
+  Widget _buildActivityEntry(BuildContext context, ActivityEntry activity,
+      ActivityEntryService service) {
+    return Column(
+      children: [
+        Container(
+          color: Color(0xFFe8d8e0),
```
Tan-jay: Personally, i would use the same colours as we have for the task_list_screen to have consistency 
like `color: index.isEven ? Colors.white : Colors.white60,`
but thats just a design decision :)

lib/services/activity_entry_service.dart
```diff
@@ -0,0 +1,58 @@
...
+      taskListID: '1',
+      timestamp: DateTime(2021, 06, 10),
+    ));
+  }
+
+  UnmodifiableListView<ActivityEntry> get activities =>
+      UnmodifiableListView(_activities);
+
+  String getCurrentDeviceName() {
+    return 'Windows Phone';
```
Tan-jay, line 47: maybe also adding a little  "ToDo" comment, that this also has to be changed with the activity sync? That it won't be forgotten :D

lib/screens/activity_log_screen.dart
```diff
@@ -0,0 +1,154 @@
...
+            Spacer(flex: 2),
+          ],
+        ),
+      );
+    }
+
+    return ListView.builder(
+      padding: EdgeInsets.zero,
+      itemCount: service.activities.length,
+      // separatorBuilder: (BuildContext context, int index) => const Divider(),
```
Tan-jay: this comment can be deleted?

lib/screens/task_list_screen.dart
```diff
@@ -124,7 +125,7 @@ class TaskListScreen extends StatelessWidget {
                 semanticLabel: 'Completed Task',
               )
             : const Icon(
-                Icons.circle_outlined,
+                EvaIcons.radioButtonOffOutline,
```
Tan-jay: You still get an error here? Probably solved with the upgrade you made, then this can be changed back :)

pubspec.lock
```diff
@@ -253,6 +253,13 @@ packages:
       url: "https://pub.dartlang.org"
     source: hosted
     version: "1.0.1"
+  eva_icons_flutter:
```
Tan-jay: this then can probably also go

pubspec.yaml
```diff
@@ -46,6 +48,7 @@ dependencies:
   # The following adds the Cupertino Icons font to your application.
   # Use with the CupertinoIcons class for iOS style icons.
   cupertino_icons: ^1.0.3
+  eva_icons_flutter: ^3.0.0
```
Tan-jay: and here :)

Tan-jay (commented):
Cool! just noticed some tiny things and if you could add those two trailing commas 😊

Tan-jay (approved):

PR #9, opened 2021-06-11
Add / adjust automated quality tools

JacobRoe:
Uses the following packages:
- dart_code_metrics
- pedantic

The package 'dart_code_metrics' adds information about code complexity, best practice recommendations etc. It's now included into the CI pipeline. The issues are shown in every PR. See https://github.com/dart-code-checker/dart-code-metrics/blob/3.x.x/doc/reporters/github-reporter.md#full-example

The package `pedantic` provides a linting rule set that is followed by every dart project of Google. The previously present issues are already fixed with this PR.

Furthermore, the coverage gets uploaded to codecov and an update to the README.md is provided on how to run the code metrics tool manually.


guisedlythe:
Please remove `network/peer.dart`, `network/socket_handler.dart` and `network/dev_playground.dart`.

Also, I am not too sure about the "Prefer trailing comma" rule, but we can keep it.

firstdayofjune:
> Also, I am not too sure about the "Prefer trailing comma" rule, but we can keep it.

I understand the trailing comma is uncommon in dart. Coming from a python background, a trailing comma feels much more natural to me. In my opinion, there are two reasons which support the trailing comma.

For one it produces smaller diffs:
> You might have noticed that closing brackets are always dedented and that a trailing comma is always added. Such formatting produces smaller diffs; when you add or remove an element, it’s always just one line.
(source: [black formatter](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html))

For another it is easy to forget the comma in the previous line, when adding another item. To me, it feels to be a common source for errors (at least I frequently find myself going back to my editor when writing JavaScript, because I forgot the comma 🙃 ). Adding the trailing comma eliminates this source of errors.

guisedlythe:
Yeah, I understand the reasoning behind it, especially for better git diffs.
I think it should be used everywhere where it is conceivable that parameters will be added/removed. Especially in function calls that can take additional parameters that are not already specified.
I don't think it makes sense to use them for function calls that already specify every argument and it is very unlikely/impossible to add/remove some of them.
I also thinks it looks weird on function declarations, but it can make sense in a lot of cases.

A point to consider (in my opinion) is, that this style increases the lines of code without adding any functionality. This can sometimes make it slower to read because you can only scan so many lines of code at once.

TLDR: I see the benefits, and we can (and will) use this style. However, there are also some arguments against it.

JacobRoe:
Odd, I could swear I chose to delete those files in the merge process. Anyway, now they are really removed. 🙄

Thank you for noticing it! 👍

Regarding the "Prefer trailing comma" rule:

Beside the arguments that Kim shared with you, I would additionally argue that on some occasions it even helps with debugging because the referring lines are specified more precise. In Java you also e.g. typically put invocations of classes that uses the builder pattern on separate lines for that reason.

I also disagree with your argument about more features in fewer lines is better or that more lines make the code slower to read. 

In fact, often times you even want to introduce additional variables to improve the readability. I also remember a case some years ago in C#, where someone wrote a parser library (for Markdown I think). One of it's main goal (or feature) was to implement a very performant one. In the end it was even comparable in speed with a C parser version. But the code ended up increasing in verticality a lot.

I can only speak for myself, but I think it makes it even easier to traverse the code. Source code is not like natural language text where visual syntax is more consistent. Meaning: Whitespace included in specific manner improves the comprehension of code because it makes it easier to distinguish different parts on a glance.

I also found this PR where someone demonstrated some corner cases where it might not make sense to use a trailing comma because how the formatter would reorganize the code in a unwanted manner (See this comment https://github.com/dart-lang/linter/pull/2566#issuecomment-816592564). But the only thing that would make my toenails crawl is the last one with skipping tests. The other ones are (imho) more readable in the trailing comma version.

guisedlythe (approved):

PR #10, opened 2021-06-12
OPTM-72 No Singleton ChangeNotifers

guisedlythe:
Refer to the commit message for more description.

OPTM-72 #close


JacobRoe (changes requested):
I see what you goal is.

Beside what I already told you in our last evening meeting, about conserving memory consumption by allowing classes to be disposed by simply scoping them to where they are needed (e.g. in Android you use Hilt for handling things like that), you could have also just use the `ChangeNotifierProvider.value` constructor to reuse provided services without them getting disposed if I'm not mistaken.

I think this change makes the access to the services in the Widget classes overly verbose for what is actually intended.

I briefly tested it with the `ChangeNotifierProvider.value` and it seems to work. Maybe you want to try it out too.

Let me know what you think.

guisedlythe:
I also the saw the `ChangeNotifierProvider.value` constructor (albeit after almost being done with this way of doing it). The reason I decided not to use it is that the introduction of this wrapper allows us to completely divide the code between stuff that is necessary for the UI (flutter) and the backend that does stuff. This will/would make it much easier to extract the backend code into a dart library that does not depend on flutter (although some of the services are only needed in conjunction with a UI, for example the network_info_service).

About the added verbosity, I think it is fine. You only see it when you use a provider, and it only affects the one line where you get the service. Considering that you should think twice before listening to a change provider anyway, I think the added wrapper is fine.

In conclusion, I would prefer this way for doing it at least for every service that would also be required in a server/library version. We could use `ChangeNotifierProvider.value` for services that are only for the UI. However, I am not sure if using two different methods to achieve the same thing in our code would only add to the confusion.

Maybe we can discuss how we proceed in our next meeting.

lib/services/identity_service.dart
```diff
@@ -32,15 +31,15 @@ class IdentityService extends ChangeNotifier with LogMixin {
 
   Future setName(String name) async {
     final updatedName = await _repository.put(_NAME_KEY, name);
-    notifyListeners();
+    if (changeCallback != null) changeCallback!();
```
JacobRoe: Is there a reason you used this line here instead of `invokeChangeCallback()`?

JacobRoe (dismissed):
I see your point.

I still think it's not the cleanest solution, especially because the ChangeCallbackNotifier is essentially dead code for most parts. But I guess it would be easy to change that later on with a real distinctive layer (aka ViewModel).

Nonetheless, I already told you that I can live with that change. I just wanted to find out if there are other options and discuss it with you. For that reason, I approved this change now and added a minor code comment down below.

lib/services/identity_service.dart
guisedlythe: There is not and I changed it.

JacobRoe (dismissed):
LGTM!

(Besides the merge conflicts.)

JacobRoe (approved):
Perfect, love it!

PR #11, opened 2021-06-13
OPTM-67 Improve QR code display

guisedlythe:
We might also want to remove the IP selection from the Settings screen.

OPTM-67 #close


lib/screens/home_screen.dart
```diff
@@ -58,4 +59,52 @@ class _HomeScreenState extends State<HomeScreen> {
...
     );
   }
+
+  void _openQrCodeDialog(BuildContext context) {
+    if (kIsWeb) return;
+    final peerService =
+        Provider.of<ChangeCallbackNotifier<PeerService>>(context, listen: false)
+            .callbackProvider;
+    if (!peerService.isServerRunning) {
+      showDialog(
```
firstdayofjune, line 69: This is a relatively long method, which moves the else statement "out of sight" from the if.
Please move the showDialog into a separate method and call this method in the if block instead.
This increases:

-  readability (because the if condition is only separated from the else by a single line)
- testability (because the method building the dialog con be tested on its own

lib/screens/home_screen.dart
```diff
@@ -58,4 +59,52 @@ class _HomeScreenState extends State<HomeScreen> {
...
+            ),
+            TextButton(
+              onPressed: () {
+                Navigator.pop(context);
+                showDialog(
+                  context: context,
+                  builder: (context) => QrCodeDialog(),
+                );
+              },
+              child: Text('No'),
```
firstdayofjune: Please change the code to close the modal, if a user selects "No".

Rational: I was confused to see the QR Code display even after selecting "No". Also, I don't see the value in displaying the QR Code without an active server.

lib/screens/qr_code_dialog.dart
```diff
@@ -35,19 +46,68 @@ class QrCodeDialog extends StatelessWidget {
...
+            final storedPort = snapshot.data![2] as int;
+
+            final ips = networkInfoService.ips;
+            final selectedIp = _selectIp(ips, storedIp);
+            if (selectedIp != storedIp && selectedIp != null) {
+              identityService.setIp(selectedIp);
+            }
+
+            return Column(children: [
+              if (!peerService.isServerRunning)
```
firstdayofjune: If the QR Code cannot be opened without a running server (see comment above), this check may be removed.

lib/screens/qr_code_dialog.dart
```diff
@@ -35,19 +46,68 @@ class QrCodeDialog extends StatelessWidget {
...
+                      TextSpan(
+                        text: ' The server is not running',
+                        style: TextStyle(fontSize: 20.0),
+                      ),
+                    ],
+                  ),
+                ),
+              Row(mainAxisAlignment: MainAxisAlignment.center, children: [
+                Text('IP: '),
+                DropdownButton<String>(
```
firstdayofjune: If there is only a single IP Address available, display the IP Address as text (like the port) instead of as a dropdown.

firstdayofjune (changes requested):
Looks good. Only some minor changes, as described in the comments.

guisedlythe:
**Regarding the alert dialog options:**
I changed the dialog options to "Yes", "Cancel", and "Show anyway" (c.f. bd7157d)
Reasoning:
I wanted to keep the option to show the QR code even if the server is not running. While you cannot connect to the device if the server is not running, you can still add it to the list of known devices. When you then start the server later, it will be used in sync. I understand that this is an edge case, but I generally dislike removing features, making it overly difficult to do what you want in some cases. I didn't add a third option before, because you can just click anywhere else on the screen to close the dialog.

firstdayofjune (approved):

PR #15, opened 2021-06-22
OPTM-55 Use own crdt library instead of the flutter crdt library

guisedlythe:
**Important**
The library is in a git submodule and integrated with a local package dependency, you need to run
```sh
git submodule init
git submodule update
```
after pulling this branch.

You also need to delete the database file (`p2p_task.db`) because the task list storage format is incompatible.

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -14,6 +14,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+        with:
+          submodules: recursive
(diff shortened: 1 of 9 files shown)


guisedlythe:
The recursive crdt will be added later in another PR.
This can be merged now so that we can use it to provide the data for the activity entries.

To get the activity entry data use the `TaskListService.taskRecords` getter, which returns an `Iterable` with tasks, their peer ids and timestamps.

guisedlythe:
Merge #16 instead

PR #16, opened 2021-06-24
OPTM-55 Recursive crdt

guisedlythe:
Preliminary implemenation of recursive Task merging using the `recursive-crdt` branch in `lww_crdt`.

There is quite some additional code in the TaskListService.


lib/models/task_list.dart
```diff
@@ -14,16 +14,18 @@ part 'task_list.g.dart';
...
   String title;
   SortOption sortBy;
-  final List<Task>? elements;
+  final List<Task> elements;
 
   TaskList({
     this.id,
     required this.title,
-    this.elements,
+    this.elements = const [],
```
firstdayofjune: This line looks dangerous to me. Is the `const []` evaluated whenever a task list is created or only at compile time? Because in the second case all TaskLists will share the same elements-list.

lib/screens/task_list_screen.dart
```diff
@@ -134,17 +135,17 @@ class _TaskListScreenState extends State<TaskListScreen> {
...
       onPressed: () {
         setState(() {
-          widget.taskList.sortBy = sortOption;
-          final listService =
-              Provider.of<ChangeCallbackNotifier<TaskListsService>>(
-            context,
-            listen: false,
-          ).callbackProvider;
-          listService.upsert(widget.taskList);
+          taskListService.upsertTaskList(
```
firstdayofjune, line 148: Sorting tasks does not work anymore. The icon next to the sort option toggles, but the tasks are not displayed in the new order.

lib/services/task_list_service.dart
```diff
@@ -23,141 +27,340 @@ class TaskListService with LogMixin, ChangeCallbackProvider {
     this._syncService,
   );
 
-  Future<List<Task>> get tasks async {
-    return (await _taskListCrdt).values;
-  }
+  Future<Iterable<TaskRecord>> get allTaskRecords async =>
+      _decodeTasks(await _readAndDecodeTaskListCollectionCrdt()).map(
+        (v) => TaskRecord(
```
firstdayofjune: what is v?

test/services/task_list_service_test.dart
```diff
@@ -1,47 +1,79 @@
...
+    expect(tasks.length, 1);
+    expect(tasks.first.title, taskTitle1);
+    final taskId = tasks.first.id!;
+    expect(taskId, isNot(null));
+
+    await device.taskListService.removeTask('id1', taskId);
+    tasks = (await device.taskListService.allTasks).toList();
+    expect(tasks.isEmpty, true);
+
+    await device.taskListService.upsertTask('id1', Task(title: taskTitle2));
```
firstdayofjune: You tested whether you can create and remove a task in the first part of the test. Why do you test the create/retrieve again? What could break by deleting a task?

test/services/task_list_service_test.dart
```diff
@@ -1,47 +1,79 @@
...
+    final task2 = Task(title: 'Drink a cold cat');
+    await device.taskListService
+        .upsertTaskList(TaskList(id: 'id1', title: 'list1'));
+    await device.taskListService.upsertTask('id1', task1);
+    await device.taskListService.upsertTask('id1', task2);
+
+    final allTaskRecords =
+        (await device.taskListService.allTaskRecords).toList();
+    expect(allTaskRecords.length, 2);
+    expect(allTaskRecords.map((v) => v.task).toSet(), {task1, task2});
```
firstdayofjune, line 82: Is the order of the set guaranteed? Could you instead assert that task1 and task2 is contained in the allTaskRecords list?

test/services/task_list_service_test.dart
```diff
@@ -1,47 +1,79 @@
...
+    expect(allTaskRecords.map((v) => v.taskListId).toList(), ['id1', 'id1']);
```
firstdayofjune, line 83: if this is possible, I would rather do something like:
```
allTaskRecords.map((record) => expect(record.taskListId, "id1"));
```

test/services/task_list_service_test.dart
```diff
@@ -1,47 +1,79 @@
...
+    const taskTitle1 = 'Catch a cat falling from the sky';
+    const taskTitle2 = 'Drink a cold cat';
+
+    await device.taskListService
+        .upsertTaskList(TaskList(id: 'id1', title: 'list1'));
+    await device.taskListService.upsertTask('id1', Task(title: taskTitle1));
+    var tasks = (await device.taskListService.allTasks).toList();
+    expect(tasks.length, 1);
+    expect(tasks.first.title, taskTitle1);
+    final taskId = tasks.first.id!;
```
firstdayofjune: Shouldn't the ID be created by the Task(...) constructor already? I would assert the ID is set in a separate test, because IMO this does not belong to the store/retrieve/delete context.

test/services/task_list_service_test.dart
```diff
@@ -1,47 +1,79 @@
...
+    final task1 = Task(title: 'Catch a cat falling from the sky');
+    await device.taskListService
+        .upsertTaskList(TaskList(id: 'id1', title: 'list1'));
+    await device.taskListService.upsertTask('id1', task1);
+    await device.taskListService.removeTask('id1', task1.id!);
+
+    final allTaskRecords =
+        (await device.taskListService.allTaskRecords).toList();
+    expect(allTaskRecords.length, 1);
+    expect(allTaskRecords.first.task, null);
```
firstdayofjune, line 96: What is the point in keeping the task as a null value in the database after removal?

lib/services/task_list_service.dart
```diff
@@ -23,141 +27,340 @@ class TaskListService with LogMixin, ChangeCallbackProvider {
...
+  Future<Iterable<TaskList>> get taskLists async => (await _taskListRecords)
+      .where((record) => !record.isDeleted)
+      .map((record) => record.value!);
+
+  Future<TaskList?> getTaskListById(String taskListId) async =>
+      (await _taskListRecordMap)[taskListId]?.value;
+
+  Future<Iterable<Task>> getTasksFromList(String taskListId) async =>
+      getTaskListById(taskListId)
+          .then((taskList) => taskList != null ? taskList.elements : []);
```
firstdayofjune, line 54: Depending on the usage, I would consider returning null for non-existing taskListIDs. The current implementation hides this error from the developer and could make debugging harder (because an existing, but empty task list behaves the same like the non-existing list).

lib/services/task_list_service.dart
```diff
@@ -23,141 +27,340 @@ class TaskListService with LogMixin, ChangeCallbackProvider {
...
+
+  Future<Iterable<Record<TaskList>>> get _taskListRecords async =>
+      (await _taskListRecordMap).values;
 
-    switch (taskList.sortBy) {
```
firstdayofjune, line 40: This was the sorting logic 😄 Did you implement this anywhere else? Otherwise this might have broken the now missing sort functionality. 🙃

lib/services/task_list_service.dart
```diff
@@ -23,141 +27,340 @@ class TaskListService with LogMixin, ChangeCallbackProvider {
...
+  /// Returns false if the task list does not exist
+  Future<bool> upsertTask(String taskListId, Task task) async {
+    l.info('Upsert task ${task.toJson()}');
+    final crdt = await _readAndDecodeTaskListCollectionCrdt();
+    final taskListCrdt = crdt.get(taskListId);
+    if (taskListCrdt == null) return false;
+
+    task.id ??= Uuid().v4();
+    final taskNode = taskListCrdt.getRecord(task.id!);
+    if (taskNode == null || taskNode.isDeleted) {
```
firstdayofjune, line 71: If this does not become to excessive, could you add a docstring explaining what a taskNode is / how the CRDT is structured? Currently, I don't understand why taskNode == null and its deletion is handled the same. (Obviously I don't understand this, because I don't know the CRDT functionality in detail. So if it is possible to summarize this in a docstring/comment would be great!)

lib/services/task_list_service.dart
```diff
@@ -23,141 +27,340 @@ class TaskListService with LogMixin, ChangeCallbackProvider {
...
-    for (var i = 0; i < allTasks.length; i++) {
-      if (allTasks[i].taskListID == taskListID) {
-        await remove(allTasks[i]);
-      }
-    }
+  Future<void> removeTask(String taskListId, String taskId) async {
+    final update = (await _readAndDecodeTaskListCollectionCrdt())
+      ..get(taskListId)?.delete(taskId);
+    await _store(update);
+    await _syncService.run();
```
firstdayofjune: Does the syncService then decide whether to actually sync or not (e.g. based on the auto-sync configuration)?

test/services/task_list_merge_test.dart
```diff
@@ -0,0 +1,189 @@
...
+  var devices = <DeviceTaskList>[];
+
+  setUp(() async {
+    devices = [
+      await DeviceTaskList.create(name: 'device1'),
+      await DeviceTaskList.create(name: 'device2'),
+    ];
+  });
+
+  test('distinct task lists', () async {
```
firstdayofjune, line 18: is this `test('multiple task lists keep their tasks distinct', ...` or test multiple task lists don't mix their tasks

test/services/task_list_merge_test.dart
```diff
@@ -0,0 +1,189 @@
...
+        .taskListService
+        .mergeCrdtJson(await devices[1].taskListService.crdtToJson());
+
+    final taskLists = (await devices[0].taskListService.taskLists).toList();
+    expect(taskLists.length, 1);
+    expect(taskLists.first.id, 'id');
+    expect(taskLists.first.title, 'list');
+    expect(taskLists.first.elements.toSet(), {task1, task2});
+    expect(
+      (await devices[0].taskListService.getTasksFromList('id')).toSet(),
```
firstdayofjune, line 66: Does toSet() guarantee the order to always be the same?

test/services/task_list_merge_test.dart
```diff
@@ -0,0 +1,189 @@
...
+      {task1, task2},
+    );
+    expect((await devices[0].taskListService.allTasks).toSet(), {task1, task2});
```
firstdayofjune, line 69: How about adding another expect to check if devices[1] also has multiple tasks now?

test/services/task_list_merge_test.dart
```diff
@@ -0,0 +1,189 @@
...
+    );
+  });
+
+  test('crdt update less recent task in list', () async {
+    final taskList = TaskList(id: 'id', title: 'list');
+    final task1 = Task(id: 'task1id', title: 'task1');
+    final task2 = Task(id: task1.id, title: 'task2');
+    await devices[0].taskListService.upsertTaskList(taskList);
+    await devices[0].taskListService.upsertTask(taskList.id!, task1);
+    await Future.delayed(Duration(milliseconds: 10));
```
firstdayofjune: Is this wait actually needed or could we instead manipulate a timestamp of a Task?

test/services/task_list_merge_test.dart
```diff
@@ -0,0 +1,189 @@
...
+    expect(
+      (await devices[0].taskListService.taskLists).toSet(),
+      {taskList1, taskList2},
+    );
+  });
+
+  test('crdt update less recent task in list', () async {
+    final taskList = TaskList(id: 'id', title: 'list');
+    final task1 = Task(id: 'task1id', title: 'task1');
+    final task2 = Task(id: task1.id, title: 'task2');
```
firstdayofjune: If there is no reason for `task1.id` please just use the same string for both tasks. This made me think. I don't want to think 😂

test/services/task_list_merge_test.dart
```diff
@@ -0,0 +1,189 @@
...
+    expect(mergedTaskList.elements.toSet(), {task2});
+  });
+
+  test('crdt keep more recent task in list', () async {
+    final taskList = TaskList(id: 'id', title: 'list');
+    final task1 = Task(id: 'task1id', title: 'task1');
+    final task2 = Task(id: task1.id, title: 'task2');
+    await devices[1].taskListService.upsertTaskList(taskList);
+    await devices[1].taskListService.upsertTask(taskList.id!, task2);
+    await Future.delayed(Duration(milliseconds: 10));
```
firstdayofjune: Same as above, can we instead set explicit timestamps? Not only does this "slow down" the tests, delays like this may also create non-reproducible test failures.

test/services/task_list_merge_test.dart
```diff
@@ -0,0 +1,189 @@
...
+
+  test('crdt recursive task merge in list', () async {
+    final taskList = TaskList(id: 'id', title: 'list');
+    final task1 = Task(
+      id: 'task1Id',
+      title: 'task1',
+      description: 'description1',
+    );
+    final task2 = Task(
+      id: task1.id,
```
firstdayofjune: Please just reuse the ID, e.g. use `"SameID"` for both tasks.

test/services/task_list_merge_test.dart
```diff
@@ -0,0 +1,189 @@
...
+        .upsertTask(taskList.id!, task1..title = 'task1 updated');
+    await devices[1]
+        .taskListService
+        .upsertTask(taskList.id!, task2..description = 'task2 description');
+
+    await devices[0]
+        .taskListService
+        .mergeCrdtJson(await devices[1].taskListService.crdtToJson());
+    expect((await devices[0].taskListService.allTasks).toSet(), {
+      Task(
```
firstdayofjune: This test can easily fail, when the Task-Constructor changes (e.g. when adding a `created` timestamp). Check for the respective values instead (so compare the description and the title explicitly).

firstdayofjune (changes requested):
Sorting functionality is now broken, except for that only some minor things 👍 
(I did not understand the CRDT implementation completely, so I did not comment anything there.)

lib/models/task_list.dart
guisedlythe: It actually creates a constant list of size 0 that can not be changed (you also cannot add new elements). Still, this can be very confusing and I will change it.

test/services/task_list_service_test.dart
guisedlythe: Deleting a task only marks the task as deleted (the task becomes a tombstone). I want to test if it can get changed back to an existing task.

test/services/task_list_service_test.dart
guisedlythe: It's a set, it doesn't have an order.
I do this to check that the tasks are contained and that they are the only tasks.

test/services/task_list_merge_test.dart
guisedlythe: Yeah, but this is easier and takes less lines of code. You actually shouldn't even need to wait because every operation takes time, this is just to be extra safe (in case it takes less than one ms). It also illustrates what this test is all about.

I get that using a delay in a test is not good style, but in this case I think that manipulating the timestamp is not worth it (and can introduce additional bugs). Also, it's just 10ms...

test/services/task_list_service_test.dart
firstdayofjune: Understood. I would split this test into several tests then (also considering my comment above).
So write one test for each:

- test task id is set after creation
- test task can be retrieved from the database
- test task can be deleted from the database (and is marked as tombstone)
- test deleted task is recovered by creating it again

To avoid copy-pasting the task creation, either create the tasks in a setup method or in a helper.
Also, you could consider grouping the tasks (https://pub.dev/packages/test#writing-tests)

test/services/task_list_service_test.dart
firstdayofjune: Ok, I wasn't sure how the comparison behaves.
Let's use the [containsAll](https://api.dart.dev/stable/2.13.4/dart-core/Set/containsAll.html) method here to make this more explicit:

`allTaskRecords.toSet().containsAll({task1, task2})`

(With allTaskRecords being a [list](https://api.dart.dev/stable/2.13.4/dart-core/List-class.html), `.toSet()` should work without the map.)

test/services/task_list_merge_test.dart
firstdayofjune: If we don't need it, remove the wait completely.

If we do, let's test the logic and make it reproducible by explicitly adding a time delta to the second task.
More lines of code is neither bad nor good. If you think, the extra lines make the test more difficult to understand, create a helper method (e.g. `incrementTimestamp(task2, Duration(milliseconds: 1));`). This is explicit and we can reuse it in other tests as well.

Adding the delay introduces a code smell to our code base, which is likely to be copied by other developers when adding further tests.

firstdayofjune (changes requested):

test/services/task_list_service_test.dart
guisedlythe: The `map` is necessary because `allTaskRecords` is of type `List<TaskRecord>`.
I think using sets for this kind of comparison is very readable and intuitive, but I agree that the previous line 58 is unnecessary and I will remove it. Using `containsAll` only checks if all elements are contained, but not if they are the only elements (in combination with line 58 it would be equivalent though).

test/services/task_list_merge_test.dart
guisedlythe: Removing it completely would, in theory, make it possible for the test to fail if it executes too quickly (less than 1ms between the insertion of both tasks).
The best I would agree to do is to change the delay to 1ms, which is the minimum for the test to be guaranteed to succeed.

**About manually adjusting the timestamp:**
It is currently not possible to  change the timestamp of tasks using the task list service because the service abstracts the underlying crdt implementation away. It is still be possible if you manipulate the json in the storage, but that is very hacky and totally dependent on the task list service implementation. This is done in the  'crdt merge clock drift' test where I even added comments that this is not a good test. I only did it there because it the only way to test this behavior on one device if you don't turn back your wall clock.
Another option would be to add a method to the task list service that allows for this timestamp manipulation. This would, however, only be useful in tests. Do you think this is worth the time investment?

firstdayofjune (approved):
Looks good! Only some final remarks (minor refactoring issues) we will cover in a separate PR as discussed.

test/services/task_list_merge_test.dart
guisedlythe: I agree that this test is not very good. It is there to test if the task lists that are created in setup() are really distinct, but it is more a test of the memory database than a test of the task list merging.

I originally added this test because I got unexpected errors at the beginning caused the databases from the different test devices not being distinct.

Should I just delete this whole test or move it into a separate file?

Solution: Move this test into separate file. Explain why it is necessary (e.g. on Windows wrong setup can lead to the use of the same db for different devices)

test/services/task_list_merge_test.dart
guisedlythe: This test only merges into device[0], so device[1] should still be unchanged.

test/services/task_list_service_test.dart
guisedlythe: Its a tombstone. The underlying implementation does not know that the task was never synced to other devices (it would not actually need a tombstone in this case).

The allTaskRecords should also return tombstones so we can display them as deleted in the activity log.

test/services/task_list_merge_test.dart
guisedlythe: => Add test to check that a task is not contained more than once in the task list.

test/services/task_list_merge_test.dart
```diff
@@ -0,0 +1,258 @@
...
+    expect(await devices[0].taskListService.getTasksFromList('id1'), [task1]);
+    expect(await devices[1].taskListService.allTasks, [task2]);
+    expect(
+      (await devices[1].taskListService.getTaskListById('id2'))?.elements,
+      [task2],
+    );
+    expect(await devices[1].taskListService.getTasksFromList('id2'), [task2]);
+  });
+
+  test('crdt merge tasks unordered same list', () async {
```
guisedlythe, line 44: Specify this (and other tests) that use one-way merge explicitly in test title.

test/services/task_list_service_test.dart
guisedlythe: See same problem as above

PR #17, opened 2021-06-25
OPTM-45 Auto-Sync

JacobRoe:
Related to:
OPTM-45, OPTM-77


lib/services/sync_service.dart
```diff
@@ -6,69 +6,78 @@ import 'package:p2p_task/utils/log_mixin.dart';
...
+  static final bool _syncOnStartDefaultValue = true;
+  static final String syncOnUpdateKey = 'syncOnUpdate';
+  static final bool _syncOnUpdateDefaultValue = true;
+
+  final KeyValueRepository _settingsRepository;
   StreamSubscription? _syncJob;
   Function()? _job;
 
-  SyncService(KeyValueRepository repository) : _repository = repository;
+  SyncService(KeyValueRepository settingsRepository)
```
guisedlythe, line 21: You can use `SyncService(this._settingsRepository);`

lib/services/sync_service.dart
```diff
@@ -6,69 +6,78 @@ import 'package:p2p_task/utils/log_mixin.dart';
...
+        await _settingsRepository.put(syncOnUpdateKey, syncOnUpdate);
     invokeChangeCallback();
 
     return updatedValue;
   }
 
-  Future startJob(Function() job) async {
+  Future<void> startJob(Function() job) async {
     _job = job;
+    if (await syncOnStart) await run();
```
guisedlythe: Should you not cancel the current job before doing the sync on start?

lib/services/sync_service.dart
```diff
@@ -6,69 +6,78 @@ import 'package:p2p_task/utils/log_mixin.dart';
...
     if (_syncJob != null) await _syncJob!.cancel();
-    final currentInterval = await interval;
-    if (currentInterval < 1) return;
-    if (await syncOnStart && currentInterval > 1) job();
     _syncJob = Stream.periodic(Duration(seconds: 1), (count) => count)
```
guisedlythe: Is it maybe possible to set the duration to the actual interval, so that we don't wake up more often than we need to?
Could complicate things when the interval changes.

guisedlythe (commented):
Can you exclude the `.mock.dart` files for the analyzer (they contain some missing comma warnings)?

JacobRoe:
> Can you exclude the `.mock.dart` files for the analyzer (they contain some missing comma warnings)?

Of course. Thanks for your comments!

Except the comment about the constructor parameter, I should've done everything you mentioned.

I also switched the `Stream` implementation one with a `Timer`-based one, which is more lightweight, because there was nobody actually listen to the `StreamSubscription`. (It also simplifies the testing of asynchronous code a little bit.)

I typically use the more verbose version without the underscore because of their semantics. The one with underscore already tells you something about the class: that the provided value is internally used as a class member. Probably a very minor thing, but I also wouldn't mind to change it. Opinions?

guisedlythe:
> I typically use the more verbose version without the underscore because of their semantics. The one with underscore already tells you something about the class: that the provided value is internally used as a class member. Probably a very minor thing, but I also wouldn't mind to change it. Opinions?

I actually agree, maybe I am using that syntactic sugar a bit too often...
However, there are use cases where it might make sense to use it even with private member variables. For example when it is clear that the constructor parameter will be stored in the class as is, but it is probably still a matter of preference.

guisedlythe (dismissed):
Looks good!

One thing I just noticed that is a bit counter-intuitive is that when 'Sync on start' is active, every change to the sync interval immediately triggers a sync.
I am not sure if we should change this though and approved this PR. If you feel that it is fine you can just merge it.

JacobRoe:
The intervals are now how they are specified on JIRA.

Also smuggled a new feature into this PR. The setting `syncAfterDeviceAdded` syncs as soon as a new device was added to the known devices. I thought it makes a great addition because the user can already add devices during the setup process. Then, when the user added at this point at least one device with existing tasks, the user would instantly see them on his own list after the setup process. A cool surprise, I think.

I also noticed a new bug in the sync service for running the sync process after specific conditions. Previously, the sync job also ran on task updates when only the option for sync on start was activated. Although I'm not sure I chose the best solution (providing flags for the run method), I would keep this approach for now and maybe add an issue.

guisedlythe (approved):
Looks good

PR #19, opened 2021-06-26
Optm 50 security

Tan-jay:
somehow the web version is super slow
auto sync works only from client to server so far
need to adapt tests


JacobRoe:
> somehow the web version is super slow
> auto sync works only from client to server so far
> need to adapt tests

Hey @Tan-jay, your work looks fabulous! 🙌 

I dived into your PR and, at least for me, it's working fine. I connected two devices via QR code. After that I deactivated every sync mechanism and created lists and tasks on both devices and synced manually. It always showed up on both devices when one device manually synced. Maybe you could elaborate on what exactly isn't working for you so that we can add a specific test for that issue.

I made some minor changes (fixing the failing tests, add a single introduction message specific test, fixing code metrics warnings, update `PeerInfo` class) but I didn't push it yet. I will clean it up later today in the evening and would push it then, if that's okay for you.

Tan-jay:
Cool, thank you! Of course totally fine with that 😊

And yeah, i think its something with my simulator... As i also have it on the main branch.. 😅 But happy that it works with you!

Tan-jay:
just added a DeletePeerMessage, that the deviceEntry is deleted on both devices when the delete button is used :)

guisedlythe:
There are currently errors in `data_model_repository_test.dart` because the `PeerInfo` constructors named parameter `status` is missing.

lib/models/peer_info.dart
```diff
@@ -23,21 +23,46 @@ class PeerLocation {
   }
 }
 
+enum Status { created, active }
```
guisedlythe, line 29: What does created and active mean? Maybe add a comment to describe what they are used for.

lib/network/messages/converter.dart
```diff
@@ -0,0 +1,20 @@
...
+
+    return Uint8List.fromList(
+      json,
+    );
+  }
+
+  @override
+  List<int> toJson(Uint8List object) {
+    return object.toList();
+  }
```
guisedlythe: Can be shortened, like:
```dart
@override
Uint8List fromJson(List<dynamic> jsonDyn) =>
    Uint8List.fromList(jsonDyn.cast<int>());

@override
List<int> toJson(Uint8List object) => object.toList();
```

lib/network/messages/introduction_message.dart
```diff
@@ -0,0 +1,35 @@
...
+  Uint8List signature;
+
+  IntroductionMessage(
+    this.peerID,
+    this.name,
+    this.ip,
+    this.port,
+    this.publicKey,
+    this.signature, {
+    this.requestReply = false,
```
guisedlythe: These are really a lot of positional parameters, maybe use named parameters instead.

lib/network/packet_handler.dart
```diff
@@ -32,7 +32,9 @@ class PacketHandler<T> {
     }
   }
 
-  void registerCallback<E>(Function(E, T source) callback) {
+  void registerCallback<E>(
+    Function(E, T source) callback,
+  ) {
```
guisedlythe: This change does nothing (probably a remnant of your development process) and can be removed, so that the file remains unchanged.

lib/network/peer/web_socket_server.dart
```diff
@@ -16,18 +17,26 @@ class WebSocketServer with LogMixin {
...
         final channelClient = WebSocketClient.fromChannel(channel);
-        final onDataCallback = createOnDataCallback(channelClient);
+        final onDataCallback = createOnDataCallback(channelClient, privateKey);
         if (onDataCallback != null) {
-          channel.stream.listen((data) => onDataCallback(channelClient, data));
+          channel.stream.listen((data) => onDataCallback(
+                channelClient,
+                data,
+                privateKey,
+              ));
```
guisedlythe: You can entirely remove the references to `privateKey` here and use closures to get it where it is needed.
Than this file can be left unchanged (except for the two added newlines below).

lib/network/web_socket_peer.dart
```diff
@@ -24,25 +26,39 @@ class WebSocketPeer with LogMixin, PacketHandler<WebSocketClient> {
...
     await _server?.close();
-    _server = await WebSocketServer.start(port, (client) {
-      logger.info('a client connected to the server');
+    _server = await WebSocketServer.start(
+      port,
+      (client, privateKey) {
+        logger.info('a client connected to the server');
 
-      return _onData;
-    });
```
guisedlythe: Like I mentioned in the previous comment, you can use a closure for this:
```dart
(client) {
  return (client, data) => _onData(client, data, privateKey);
}
```

lib/network/web_socket_peer.dart
```diff
@@ -24,25 +26,39 @@ class WebSocketPeer with LogMixin, PacketHandler<WebSocketClient> {
...
-    logger.info('Received message from connected peer: $data');
-    _handleMessage(client, data);
+  void _onData(
+    WebSocketClient client,
+    dynamic data,
+    RSAPrivateKey? privateKey,
+  ) {
+    logger.info('Received message from connected peer');
+    _handleMessage(client, data, privateKey);
   }
```
guisedlythe: This method doesn't seem to do anything but forward the function call to another function. Maybe remove it?

lib/network/web_socket_peer.dart
```diff
@@ -137,7 +169,9 @@ class WebSocketPeer with LogMixin, PacketHandler<WebSocketClient> {
     return await completer.future;
   }
 
-  WebSocketClient? tryWebSocketClientConnect(Uri uri) {
+  WebSocketClient? tryWebSocketClientConnect(
+    Uri uri,
+  ) {
```
guisedlythe: Probably don't need the trailing comma in this case. You can leave it though if you want.

lib/screens/devices/device_form_screen.dart
```diff
@@ -56,6 +62,53 @@ class _DeviceFormScreenState extends State<DeviceFormScreen> {
...
+                    if (value == null || value.isEmpty) {
+                      return 'Give the device an id.';
+                    }
+
+                    return null;
+                  },
+                  onFieldSubmitted: (value) {
+                    _idFocusNode.requestFocus();
+                  },
+                ),
```
guisedlythe, line 85: Why do we need a name and an ID? Shouldn't the actual device ID be sent from the device.

lib/screens/devices/device_form_screen.dart
```diff
@@ -56,6 +62,53 @@ class _DeviceFormScreenState extends State<DeviceFormScreen> {
...
+                    if (value == null || value.isEmpty) {
+                      return 'The publicKey address is missing.';
+                    }
+
+                    return null;
+                  },
+                  onFieldSubmitted: (value) {
+                    _publicKeyFocusNode.requestFocus();
+                  },
+                ),
```
guisedlythe: Can we change this text field to support multiple lines?
Not having newlines might actually butcher the key structure (I haven't really tried it yet though).

lib/screens/home_screen.dart
```diff
@@ -33,8 +34,10 @@ class _HomeScreenState extends State<HomeScreen> {
       context,
       listen: false,
     ).callbackProvider;
+
+    _handleEncryptionKeys(identityService);
     identityService.name.then((value) {
-      if (value == null) {
+      if (value == '') {
```
guisedlythe: Can use `value.isEmpty`

lib/screens/home_screen.dart
```diff
@@ -45,6 +48,25 @@ class _HomeScreenState extends State<HomeScreen> {
...
+      await identityService.setPrivateKeyPem(privatekeyPem);
+      await identityService.setPublicKeyPem(publicKeyPem);
+    } else {
+      // print public key for manual adding
+      // new line has to be replaced manually by \r\n
+      // -----BEGIN RSA PUBLIC KEY-----
+      // MIIFog...
+      // -----END RSA PUBLIC KEY-----
+      // => -----BEGIN RSA PUBLIC KEY-----\r\nMIIFog...\r\n-----BEGIN RSA PUBLIC KEY-----
+      print(await identityService.publicKeyPem);
```
guisedlythe: Probably related to my previous comment about the single-line public key text field when adding peers manually.

lib/screens/qr_code_dialog.dart
```diff
@@ -178,8 +198,16 @@ class QrCodeDialog extends StatelessWidget with LogMixin {
...
 
-  String _makeQrContent(String deviceName, String ip, int port) =>
-      '$deviceName,$ip,$port';
+  // ignore: long-parameter-list
+  String _makeQrContent(
+    String name,
+    String ip,
+    int port,
+    String publicKey,
+    String peerID,
```
guisedlythe, line 207: Maybe also use named parameters here.
Do we really need to have `name` and `peerId` in the QR code? If we really need them we should consider encoding them (e.g. base64) so that they cannot interfere with our very primitive `peerId,name,ip,...` QR text encoding.

lib/screens/qr_code_dialog.dart
```diff
@@ -197,6 +225,15 @@ class _ConnectionInfo {
...
+  final String peerID;
+  final String publicKey;
+
+  const _ConnectionInfo(
+    this.selectedIp,
+    this.ips,
+    this.port,
+    this.deviceName,
+    this.peerID,
+    this.publicKey,
```
guisedlythe: Another case for named parameters

lib/screens/settings/database_list_section.dart
```diff
@@ -21,7 +21,7 @@ class DatabaseSection extends StatelessWidget {
 
     return FutureBuilder<int>(
       initialData: -1,
-      future: taskListService.count(),
+      future: taskListService.countLists(),
```
guisedlythe: I already changed this in #32 when we started to refactor the task list service. Maybe just reset this file and leave it to another PR. It doesn't really have anything to do security.

lib/screens/settings/identity_list_section.dart
```diff
@@ -82,6 +82,25 @@ class IdentityListSection extends StatelessWidget {
...
+              );
+            }
+            final publicKey = snapshot.data!;
+
+            return ListTile(
+              tileColor: Colors.white,
+              leading: Icon(Icons.vpn_key),
+              title: Text('Public key'),
+              subtitle: Text(publicKey),
+            );
```
guisedlythe: Is it maybe possible to not show the public key in the subtitle but to open a dialog when you click on the tile that displays the public key? Another very cool feature would be a button to copy it to the clipboard (could be done in the dialog), especially since I can't seem to copy it in the desktop version for some reason.

lib/security/key_helper.dart
```diff
@@ -0,0 +1,252 @@
...
+      Uint8List.fromList(
+        utf8.encode(dataToEncrypt),
+      ),
+    );
+
+    return String.fromCharCodes(blocks);
+  }
+
+  Uint8List rsaSign(pc.RSAPrivateKey privateKey, String dataToSign) {
+    final signer = pc.RSASigner(pc.SHA256Digest(), '0609608648016503040201');
```
guisedlythe: Maybe add a member constant for the value of the string. The constant name would than also be a hint to what this is actually used for.

lib/security/key_helper.dart
```diff
@@ -0,0 +1,252 @@
...
+
+    return sig.bytes;
+  }
+
+  bool rsaVerify(String publicKeyPem, String signedData, Uint8List signature) {
+    var publicKey = decodePublicKeyFromPem(publicKeyPem);
+
+    final sig = pc.RSASignature(signature);
+
+    final verifier = pc.RSASigner(pc.SHA256Digest(), '0609608648016503040201');
```
guisedlythe: Same as the previous comment

lib/security/key_helper.dart
```diff
@@ -0,0 +1,252 @@
...
+
+    verifier.init(
+      false,
+      pc.PublicKeyParameter<pc.RSAPublicKey>(publicKey),
+    ); // false=verify
```
guisedlythe: What does this comment mean?

lib/security/key_helper.dart
```diff
@@ -0,0 +1,252 @@
...
+      ),
+    );
+
+    return secureRandom;
+  }
+
+  pc.RSAPrivateKey decodePrivateKeyFromPem(String priavateKeyPem) {
+    try {
+      var key = encrypt_pem.RSAKeyParser()
+          .parse(priavateKeyPem.replaceAll('\\r\\n', '\n')) as pc.RSAPrivateKey;
```
guisedlythe: Could probably add a function for this since this line appears twice.

lib/security/key_helper.dart
```diff
@@ -0,0 +1,252 @@
+import 'dart:math';
```
guisedlythe, line 1: There are a lot of `var` variables that can be made `final`

lib/security/key_helper.dart
```diff
@@ -0,0 +1,252 @@
...
+    topLevel.add(
+      ASN1Integer(publicKey.modulus!),
+    );
+    topLevel.add(
+      ASN1Integer(publicKey.exponent!),
+    );
+
+    var dataBase64 = base64.encode(topLevel.encodedBytes);
+
+    return '-----BEGIN RSA PUBLIC KEY-----\r\n$dataBase64\r\n-----END RSA PUBLIC KEY-----';
```
guisedlythe: Also maybe add a function to wrap the key in the 'BEGIN ... END' since it is done more than once and largely contains string constants.

lib/services/identity_service.dart
```diff
@@ -37,6 +41,50 @@ class IdentityService with LogMixin, ChangeCallbackProvider {
...
+  Future<RSAPublicKey?> get publicKey async {
+    if (await publicKeyPem == '') {
+      return null;
+    }
+
+    try {
+      return KeyHelper().decodePublicKeyFromPem(await publicKeyPem);
+    } on Exception catch (error) {
+      logger.severe('could not parse privateKey $error');
+    }
```
guisedlythe: Should store the `await publicKeyPem` in a variable that is doesn't have to be done twice.

lib/services/identity_service.dart
```diff
@@ -37,6 +41,50 @@ class IdentityService with LogMixin, ChangeCallbackProvider {
...
+  Future<RSAPrivateKey?> get privateKey async {
+    if (await privateKeyPem == '') {
+      return null;
+    }
+
+    try {
+      return KeyHelper().decodePrivateKeyFromPem(await privateKeyPem);
+    } on Exception catch (error) {
+      logger.severe('could not parse privateKey $error');
+    }
```
guisedlythe: Same as previous comment

lib/services/peer_info_service.dart
```diff
@@ -11,6 +11,19 @@ class PeerInfoService with ChangeCallbackProvider {
...
+
+    for (var i = 0; i < allDevices.length; i++) {
+      if (allDevices[i].status == Status.active) {
+        activeDevices.add(allDevices[i]);
+      }
+    }
+
+    return activeDevices;
+  }
+
```
guisedlythe: Can be shortened to just:
```dart
Future<List<PeerInfo>> get activeDevices async => (await devices)
    .where((device) => device.status == Status.active)
    .toList();
```

lib/services/peer_info_service.dart
```diff
@@ -22,6 +35,17 @@ class PeerInfoService with ChangeCallbackProvider {
...
+    var allDevices = await devices;
+    for (var i = 0; i < allDevices.length; i++) {
+      if (allDevices[i].id == id) {
+        return allDevices[i];
+      }
+    }
+
+    return null;
+  }
+
```
guisedlythe: Can also be shortened (even though dart streams are a bit weird when finding stuff with null safety and this is actually pretty ugly):
```dart
try {
  return (await devices).firstWhere((device) => device.id == id);
} on StateError {
  return null;
}
```

lib/services/task_list_service.dart
```diff
@@ -141,6 +141,10 @@ class TaskListService with LogMixin, ChangeCallbackProvider {
     return (await allTasks).length;
   }
 
+  Future<int> countLists() async {
+    return (await taskLists).length;
+  }
+
```
guisedlythe: As mentioned before, this should probably be removed from this PR.

lib/utils/key_value_repository.dart
```diff
@@ -18,7 +18,11 @@ class KeyValueRepository with LogMixin {
 
   /// [value] must have a type that is supported by sembast
   Future<dynamic> put(String key, dynamic value) async {
-    logger.info('Put setting with key "$key" and value "$value"');
+    if (key == 'privateKeyKey') {
+      logger.info('Put setting with key "$key"');
+    } else {
+      logger.info('Put setting with key "$key" and value "$value"');
+    }
```
guisedlythe: What is the meaning of this special string value that only changes the log message?

test/network/web_socket_peer_test.dart
```diff
@@ -51,23 +81,74 @@ void main() {
...
+      expect(currentPeerInfo, equals(null));
+
+      await devices[0].peerService.sendIntroductionMessageToPeer(
+            peerInfoDevice1,
+            location: peerInfoDevice1.locations.first,
+          );
+
+      final newPeerInfo =
+          await devices[1].peerInfoService.getByID(peerInfoDevice0.id!);
+      expect(newPeerInfo, isNot(equals(null)));
```
guisedlythe: `isNot(null)` should suffice, I don't think the `equals` is needed (also in later tests)

test/security/security_test.dart
```diff
@@ -0,0 +1,84 @@
...
+  var keyPair = keyHelper.generateRSAkeyPair();
+  var message = 'Really private stuff';
+
+  group('#security', () {
+    test('encryption with keys', () {
+      var encrypted = keyHelper.encrypt(keyPair.publicKey, message);
+      expect(message != encrypted, true);
+      var plain = keyHelper.decrypt(keyPair.privateKey, encrypted);
+
+      expect(message == plain, true);
```
guisedlythe: Should be possible to use `expect(message, isNot(plain))` to produce better error messages if it fails (also in some other tests).

guisedlythe (commented):
This is a only preliminary review and I will go through it in more detail tomorrow (actually today). I mostly only went through the files on github and did not look too much into how everything comes together and works yet, I also didn't do any extensive testing yet.

A couple of simple tests worked though, which is pretty damn cool!

One thing I noticed is that almost every `var` can be changed to `final`.

lib/screens/devices/device_form_screen.dart
Tan-jay: For each send message, the peerID of the sending device is used to create a signature. So therefore i need to have the peerID in advance to proof if the signature is valid (in this case it is used in the _handleIntroductionReplyMessage to set the peer to active)

lib/screens/qr_code_dialog.dart
Tan-jay: yes peerID is really important for the security part and it's weird to not have a name set under the device tab to identify. 
What do you mean bei interfere, like how?

lib/screens/settings/identity_list_section.dart
Tan-jay: also added an option to copy the peerID

lib/security/key_helper.dart
Tan-jay: basically
true = signing
false = verifying

but yeah i removed it, i guess its fine to check the parameter name

lib/security/key_helper.dart
Tan-jay: its one time with public and one time with private key, but made it a bit more readable :)

lib/utils/key_value_repository.dart
Tan-jay: just for security the private key should not be printed without a good reason and it spams the logs as it is quite big

test/security/security_test.dart
Tan-jay: expect(message, equals(plain)) 👌

lib/utils/key_value_repository.dart
guisedlythe: In that case I suggest adding an optional named parameter to the put function which can be used to disable logging the value. Checking for a string constant like this adds a really weird and hard to maintain dependency from `KeyValueRepository` to `IdentityService`.

test/security/security_test.dart
guisedlythe: Nice catch but I originally meant line 17 :smile:

lib/screens/devices/device_form_screen.dart
guisedlythe: Hmm, but shouldn't this be unnecessary because the owner of the private key is the only one who can decrypt messages that are encrypted with its public key. He is also the only one who can encrypt the message so that we decrypt it with their public key. We should only need to make sure that the encryption/decryption was successful, so just sending a hard-coded encrypted hello message and checking if it was received correctly after decryption should suffice, or am I missing something?

We could then send the name and the peer ID in the introduction messages and fill them in at this point. The only downside is that we would not know the peerId and peerName until we have actually communicated with the other party.

Tan-jay: Actually there are 2 problems if we don't set the peerID:

1. If the introduction reply messages comes back (check for success), we don't know to which peer it belongs -> I cannot connect the publicKey with the peerID. Imagine you add 2 devices in the beginning and all servers are not activated, then we have 2 open peer connections and when we then later retry to send the introduction message and a reply comes, we don't know from which one this came. 

2. (considering that we solved the matching problem): An attacker could try to manipulate the connection to always fail
  A -> B: introduction message which failed 
  C -> A: send introduction reply message (as A's public key and location is public - imagine random tries)
   => A thinks it worked and will send messages, B doesn't know A and will always decline messages.

Was this understandable? 😃

And having in mind that this is more a testing "issue" as usually you would scan the QR code (the server is active -> always success) and don't have to deal with this, i think it's okay to have the two extra fields.

test/security/security_test.dart
Tan-jay: oh yes haha 😄

lib/screens/devices/device_form_screen.dart
guisedlythe: Thanks, I think I understand the problem.

lib/screens/qr_code_dialog.dart
guisedlythe: If the name, contains a comma (e.g. `this,is,the,name`), the QR content decryption will fail. But I think there is already a bug for that in JIRA and we can do it in a future PR (OPTM-73).

lib/utils/key_value_repository.dart
```diff
@@ -17,11 +17,11 @@ class KeyValueRepository with LogMixin {
...
   /// [value] must have a type that is supported by sembast
-  Future<dynamic> put(String key, dynamic value) async {
-    if (key == 'privateKeyKey') {
-      logger.info('Put setting with key "$key"');
-    } else {
+  Future<dynamic> put(String key, dynamic value, bool displayValue) async {
+    if (displayValue) {
       logger.info('Put setting with key "$key" and value "$value"');
+    } else {
+      logger.info('Put setting with key "$key"');
```
guisedlythe: I was thinking more like:
```dart
Future<dynamic> put(String key, dynamic value, {bool displayValue = true,}) async { ... }
```
This way we don't need to change the old code and only specify it `displayValue: false` where it is necessary.

lib/viewmodels/device_list_viewmodel.dart
```diff
@@ -35,47 +35,80 @@ class DeviceListViewModel with LogMixin {
...
   }
 
   void remove(PeerInfo peer) async {
+    try {
+      await _peerService.sendDeletePeerMessageToPeer(peer);
+    } on FormatException catch (e) {
+      logger.warning('could not send delete peer message - $e');
+    }
+
     await _peerInfoService.remove(peer);
```
guisedlythe: This has the weird effect that you cannot remove a device locally if the message cannot be sent (i.e. there is another error in the message sending function). This can happen very easily, for example if you just entered the wrong connection info. I suggest that is always possible to delete peers, even if the message cannot be sent (maybe display a warning and a confirm dialog).

Maybe use something like
```dart
try {
  ...
} finally {
  await _peerInfoService.remove(peer);
}
```

I encountered this when the desktop version added the web version as device with an empty `ip` field because web versions cannot start servers. Devices that cannot be connected to should probably be handled and displayed differently (we do need them for their keys though).

lib/utils/key_value_repository.dart
Tan-jay: oh yes 😄

guisedlythe (commented):
I have tried some things with my desktop connecting to my android phone and it worked fine.

There is one issue that the device list does not refresh when devices are removed/added/updated. This is because the device view model does not actually listen to changes from the peer info service. I don't really know how to do that though because `Provider.of` cannot easily be used in the view model class (I think).

I think the biggest issue is that this makes the web version pretty much unusable because it is suuuper slow. I found some related issues to that and I suggest that we create an issue for it to improve the web experience later:
https://github.com/bcgit/pc-dart/issues/113 (web key-gen slow)
https://github.com/bcgit/pc-dart/issues/87 (not about web but general platform deoendent performance improvements)

There are some minor changes (mostly typos) that I did in my local files. You could just apply them (`git apply`):
```diff
diff --git a/lib/models/peer_info.dart b/lib/models/peer_info.dart
index 7da38a6..8029a94 100644
--- a/lib/models/peer_info.dart
+++ b/lib/models/peer_info.dart
@@ -23,9 +23,9 @@ class PeerLocation {
   }
 }
 
-// Status created indicates that the key exchange has started, but not finished yet
-// it might has failed becasue the sever was off.
-// when it was successfull it tunres to active.
+// Status created indicates that the key exchange has started, but not finished yet.
+// It might have failed because the sever was off.
+// Once it is successfull, the status will be set to active.
 enum Status { created, active }
 
 @JsonSerializable(explicitToJson: true)
diff --git a/lib/screens/home_screen.dart b/lib/screens/home_screen.dart
index 95c9775..65e291c 100644
--- a/lib/screens/home_screen.dart
+++ b/lib/screens/home_screen.dart
@@ -49,9 +49,9 @@ class _HomeScreenState extends State<HomeScreen> {
   }
 
   void _handleEncryptionKeys(IdentityService identityService) async {
-    if (await identityService.publicKeyPem == '') {
+    if ((await identityService.publicKeyPem).isEmpty) {
       var keyHelper = KeyHelper();
-      var pair = keyHelper.generateRSAkeyPair();
+      var pair = keyHelper.generateRSAKeyPair();
       var privatekeyPem = keyHelper.encodePrivateKeyToPem(pair.privateKey);
       var publicKeyPem = keyHelper.encodePublicKeyToPem(pair.publicKey);
       await identityService.setPrivateKeyPem(privatekeyPem);
diff --git a/lib/security/key_helper.dart b/lib/security/key_helper.dart
index 2cd66fe..680b682 100644
--- a/lib/security/key_helper.dart
+++ b/lib/security/key_helper.dart
@@ -1,20 +1,20 @@
 import 'dart:math';
 import 'dart:typed_data';
 import 'dart:convert';
+import 'package:p2p_task/utils/log_mixin.dart';
 import 'package:pointycastle/export.dart' as pc;
 import 'package:asn1lib/asn1lib.dart';
 import 'package:encrypt/encrypt.dart' as encrypt_pem;
 
-class KeyHelper {
-  final algorithmIdentifier = '0609608648016503040201';
+class KeyHelper with LogMixin {
+  static const _algorithmIdentifier = '0609608648016503040201';
 
-  pc.AsymmetricKeyPair<pc.RSAPublicKey, pc.RSAPrivateKey> generateRSAkeyPair({
+  pc.AsymmetricKeyPair<pc.RSAPublicKey, pc.RSAPrivateKey> generateRSAKeyPair({
     int bitLength = 2048,
   }) {
     // higher value of bitLength will lower the performance
 
     final keyGen = pc.KeyGenerator('RSA');
-
     keyGen.init(
       pc.ParametersWithRandom(
         pc.RSAKeyGeneratorParameters(BigInt.parse('65537'), bitLength, 64),
@@ -22,7 +22,12 @@ class KeyHelper {
       ),
     );
 
+    logger.info('Generating key pair');
+    final start = DateTime.now();
     final pair = keyGen.generateKeyPair();
+    final end = DateTime.now();
+    logger.info('Generating key pair (took ${end.difference(start)})');
+
     final myPublic = pair.publicKey as pc.RSAPublicKey;
     final myPrivate = pair.privateKey as pc.RSAPrivateKey;
 
@@ -110,7 +115,7 @@ class KeyHelper {
   }
 
   Uint8List rsaSign(pc.RSAPrivateKey privateKey, String dataToSign) {
-    final signer = pc.RSASigner(pc.SHA256Digest(), algorithmIdentifier);
+    final signer = pc.RSASigner(pc.SHA256Digest(), _algorithmIdentifier);
 
     signer.init(true, pc.PrivateKeyParameter<pc.RSAPrivateKey>(privateKey));
 
@@ -126,7 +131,7 @@ class KeyHelper {
 
     final sig = pc.RSASignature(signature);
 
-    final verifier = pc.RSASigner(pc.SHA256Digest(), algorithmIdentifier);
+    final verifier = pc.RSASigner(pc.SHA256Digest(), _algorithmIdentifier);
 
     verifier.init(
       false,
@@ -162,14 +167,14 @@ class KeyHelper {
     return secureRandom;
   }
 
-  pc.RSAPrivateKey decodePrivateKeyFromPem(String priavateKeyPem) {
+  pc.RSAPrivateKey decodePrivateKeyFromPem(String privateKeyPem) {
     try {
-      final key = encrypt_pem.RSAKeyParser().parse(_sanitizePem(priavateKeyPem))
+      final key = encrypt_pem.RSAKeyParser().parse(_sanitizePem(privateKeyPem))
           as pc.RSAPrivateKey;
 
       return key;
     } on FormatException {
-      return encrypt_pem.RSAKeyParser().parse(priavateKeyPem)
+      return encrypt_pem.RSAKeyParser().parse(privateKeyPem)
           as pc.RSAPrivateKey;
     }
   }
diff --git a/lib/services/identity_service.dart b/lib/services/identity_service.dart
index 02ddf88..25e1834 100644
--- a/lib/services/identity_service.dart
+++ b/lib/services/identity_service.dart
@@ -46,10 +46,7 @@ class IdentityService with LogMixin, ChangeCallbackProvider {
 
   Future<RSAPublicKey?> get publicKey async {
     final _publicKeyPem = await publicKeyPem;
-
-    if (_publicKeyPem == '') {
-      return null;
-    }
+    if (_publicKeyPem.isEmpty) return null;
 
     try {
       return KeyHelper().decodePublicKeyFromPem(_publicKeyPem);
@@ -70,10 +67,7 @@ class IdentityService with LogMixin, ChangeCallbackProvider {
 
   Future<RSAPrivateKey?> get privateKey async {
     final _privateKeyPem = await privateKeyPem;
-
-    if (_privateKeyPem == '') {
-      return null;
-    }
+    if (_privateKeyPem.isEmpty) return null;
 
     try {
       return KeyHelper().decodePrivateKeyFromPem(_privateKeyPem);
diff --git a/lib/services/peer_service.dart b/lib/services/peer_service.dart
index 6f19857..50cd62a 100644
--- a/lib/services/peer_service.dart
+++ b/lib/services/peer_service.dart
@@ -91,7 +91,7 @@ class PeerService with LogMixin, ChangeCallbackProvider {
       deletePeerMessage.signature,
     )) {
       logger
-          .warning('Cannot verify signaure of delete peer message - skipping');
+          .warning('Cannot verify signature of delete peer message - skipping');
 
       return;
     }
diff --git a/lib/utils/key_value_repository.dart b/lib/utils/key_value_repository.dart
index f0fb2f3..a8671aa 100644
--- a/lib/utils/key_value_repository.dart
+++ b/lib/utils/key_value_repository.dart
@@ -17,8 +17,11 @@ class KeyValueRepository with LogMixin {
   }
 
   /// [value] must have a type that is supported by sembast
-  Future<dynamic> put(String key, dynamic value,
-      {bool displayValue = true}) async {
+  Future<dynamic> put(
+    String key,
+    dynamic value, {
+    bool displayValue = true,
+  }) async {
     if (displayValue) {
       logger.info('Put setting with key "$key" and value "$value"');
     } else {
diff --git a/test/security/security_test.dart b/test/security/security_test.dart
index e8bda75..17c5c3c 100644
--- a/test/security/security_test.dart
+++ b/test/security/security_test.dart
@@ -8,7 +8,7 @@ import 'package:p2p_task/network/web_socket_peer.dart';
 
 void main() {
   var keyHelper = KeyHelper();
-  var keyPair = keyHelper.generateRSAkeyPair();
+  var keyPair = keyHelper.generateRSAKeyPair();
   var message = 'Really private stuff';
 
   group('#security', () {
@@ -16,8 +16,7 @@ void main() {
       var encrypted = keyHelper.encrypt(keyPair.publicKey, message);
       expect(message, isNot(encrypted));
       var plain = keyHelper.decrypt(keyPair.privateKey, encrypted);
-
-      expect(message, equals(plain));
+      expect(message, plain);
     });
 
     test('encryption with pem files', () {
@@ -27,22 +26,19 @@ void main() {
       var encrypted = keyHelper.encryptWithPublicKeyPem(publicKeyPem, message);
       expect(message, isNot(encrypted));
       var plain = keyHelper.decryptWithPrivateKeyPem(privateKeyPem, encrypted);
-
-      expect(message, equals(plain));
+      expect(message, plain);
     });
 
     test('sign and verify message', () {
-      var publicKeyPem = keyHelper.encodePublicKeyToPem(keyPair.publicKey);
-      var signature = keyHelper.rsaSign(keyPair.privateKey, message);
+      final publicKeyPem = keyHelper.encodePublicKeyToPem(keyPair.publicKey);
+      final signature = keyHelper.rsaSign(keyPair.privateKey, message);
       expect(keyHelper.rsaVerify(publicKeyPem, message, signature), true);
     });
 
-    test('check signature of decrpyted message', () {
-      var publicKeyPem = keyHelper.encodePublicKeyToPem(keyPair.publicKey);
-
-      var sig = keyHelper.rsaSign(keyPair.privateKey, 'peerID');
-
-      var introductionMessage = IntroductionMessage(
+    test('check signature of decrypted message', () {
+      final publicKeyPem = keyHelper.encodePublicKeyToPem(keyPair.publicKey);
+      final sig = keyHelper.rsaSign(keyPair.privateKey, 'peerID');
+      final introductionMessage = IntroductionMessage(
         peerID: 'peerID',
         name: 'name',
         ip: 'ip',
@@ -52,21 +48,17 @@ void main() {
       );
 
       final peer = WebSocketPeer();
-
       peer.registerTypename<IntroductionMessage>(
         'IntroductionMessage',
         (json) => IntroductionMessage.fromJson(json),
       );
 
-      var packet = peer.marshallPacket(introductionMessage);
-      var encrypted = keyHelper.encryptWithPublicKeyPem(publicKeyPem, packet);
-
-      expect(encrypted != packet, true);
-
-      var decrypted = keyHelper.decrypt(keyPair.privateKey, encrypted);
-      var decryptedPacket = Packet.fromJson(jsonDecode(decrypted));
-
-      var plainMessage = IntroductionMessage.fromJson(decryptedPacket.object);
+      final packet = peer.marshallPacket(introductionMessage);
+      final encrypted = keyHelper.encryptWithPublicKeyPem(publicKeyPem, packet);
+      expect(encrypted, isNot(packet));
+      final decrypted = keyHelper.decrypt(keyPair.privateKey, encrypted);
+      final decryptedPacket = Packet.fromJson(jsonDecode(decrypted));
+      final plainMessage = IntroductionMessage.fromJson(decryptedPacket.object);
 
       expect(
         keyHelper.rsaVerify(
diff --git a/test/utils/device_task_list.dart b/test/utils/device_task_list.dart
index 2b1c8c8..118fdb9 100644
--- a/test/utils/device_task_list.dart
+++ b/test/utils/device_task_list.dart
@@ -42,7 +42,7 @@ class DeviceTaskList {
 
   static Future<void> _createKeys(IdentityService identityService) async {
     final keyHelper = KeyHelper();
-    final keys = keyHelper.generateRSAkeyPair();
+    final keys = keyHelper.generateRSAKeyPair();
     final privateKey = keyHelper.encodePrivateKeyToPem(keys.privateKey);
     final publicKey = keyHelper.encodePublicKeyToPem(keys.publicKey);
     await identityService.setPrivateKeyPem(privateKey);
diff --git a/test/websocket_test.dart b/test/websocket_test.dart
index 15e6cd4..eda4ee3 100644
--- a/test/websocket_test.dart
+++ b/test/websocket_test.dart
@@ -23,7 +23,7 @@ void main() {
     await webSocket.close();
     final value = await completer.future;
 
-    expect(value, equals('Hello, world!'));
+    expect(value, 'Hello, world!');
   });
 
   test('connects to a custom-made WebSocketPeer', () async {
@@ -31,7 +31,7 @@ void main() {
     var receivePort = ReceivePort();
 
     final keyHelper = KeyHelper();
-    final keys = keyHelper.generateRSAkeyPair();
+    final keys = keyHelper.generateRSAKeyPair();
     final publicKey = keyHelper.encodePublicKeyToPem(keys.publicKey);
     await Isolate.spawn(
       startServer,
```

guisedlythe (approved):
Looks good and should be ready for merge!

PR #20, opened 2021-06-26
Database related features

JacobRoe:
Related to:
OPTM-56, OPTM-57, OPTM-60, OPTM-69, OPTM-81


lib/config/style_constants.dart
```diff
@@ -2,3 +2,7 @@ import 'package:flutter/material.dart';
 
 const kGrayBackground = Color(0xFFEFEFF4);
 const kHeroFont = TextStyle(fontSize: 24.0, fontWeight: FontWeight.bold);
+final kStandardTheme = ThemeData(
+  primarySwatch: Colors.deepPurple,
+  canvasColor: kGrayBackground,
+);
```
guisedlythe: A bit weird that this is not an actual constant, but I don't know how it could be changed for the better.

lib/screens/activity_log_screen.dart
```diff
@@ -1,4 +1,5 @@
 import 'package:flutter/material.dart';
```
guisedlythe, line 1: I understand the changes (they are definitely good changes), but this has nothing to do with what this PR wants to achieve.
Please add these kind of changes in other PRs because it makes the review and merge process much harder.

lib/screens/activity_log_screen.dart
```diff
@@ -111,13 +110,16 @@ class ActivityLogScreen extends StatelessWidget {
...
+    return Icon(activity.device == service.getCurrentDeviceName()
+        ? Icons.arrow_upward
+        : Icons.arrow_downward);
+    // return Image.asset(
+    //   activity.device == service.getCurrentDeviceName()
+    //       ? 'assets/up_arrow_icon.png'
+    //       : 'assets/down_arrow_icon.png',
+    //   width: 22,
+    //   height: 22,
+    // );
```
guisedlythe: But while you are already on it, you can remove this comment with the old code.

lib/screens/devices/device_list_screen.dart
```diff
@@ -19,118 +16,128 @@ class DeviceListScreen extends StatefulWidget {
...
+          viewModel,
         ),
+        _buildFloatingButton(context, viewModel),
       ],
     );
   }
 
-  Widget _buildDeviceList(
+  Widget _buildFloatingButton(
+      BuildContext context, DeviceListViewModel viewModel) {
```
guisedlythe: Prefer trailing comma.

lib/screens/devices/device_list_screen.dart
```diff
@@ -5,10 +5,7 @@ import 'package:p2p_task/config/style_constants.dart';
 import 'package:p2p_task/models/peer_info.dart';
```
guisedlythe, line 5: Good change using a viewmodel to separate the logic from the UI, but also not related to the database...
In theory I would need to test the entire device list screen again (but it appeared to work at first glance and I trust everything is in order).

lib/screens/settings/database_list_section.dart
```diff
@@ -57,9 +64,39 @@ class DatabaseSection extends StatelessWidget {
...
+              title: Text('Reset database'),
+              subtitle: Text(databaseService.database?.path ?? ''),
+            ),
           ],
         );
       },
     );
   }
+
+  void handleDatabaseDeletion(
```
guisedlythe: Can be private.

lib/screens/settings/database_list_section.dart
```diff
@@ -57,9 +64,39 @@ class DatabaseSection extends StatelessWidget {
...
+      context,
+      listen: false,
+    ).callbackProvider;
+    final confirmed =
+        await YesNoDialog.show(context, title: 'Delete all data?') ?? false;
+    if (confirmed) {
+      await databaseService.delete();
+      await sharedPreferences.clear();
+      await syncService.clearJob();
+      DependenciesProvider.rebuild(context);
```
guisedlythe: Either ask for device name after setting reset or keep the old device name. Currently the device name just becomes null.

lib/screens/settings/settings_screen.dart
```diff
@@ -3,17 +3,51 @@ import 'package:p2p_task/screens/settings/database_list_section.dart';
 import 'package:p2p_task/screens/settings/identity_list_section.dart';
 import 'package:p2p_task/screens/settings/network_list_section.dart';
 import 'package:p2p_task/screens/settings/sync_list_section.dart';
+import 'package:p2p_task/services/database_service.dart';
+import 'package:p2p_task/utils/store_ref_names.dart';
+import 'package:p2p_task/widgets/yes_no_dialog.dart';
+import 'package:provider/provider.dart';
 
-class SettingsScreen extends StatelessWidget {
+class SettingsScreen extends StatefulWidget {
```
guisedlythe, line 15: Is there a reason for the settings screen to become a `StatefulWidget`?

lib/screens/setup/database_config_screen.dart
```diff
@@ -0,0 +1,113 @@
...
+              helperMaxLines: 2,
+              helperText:
+                  'The task lists and tasks you create, will be stored in a file at this location.',
+            ),
+            controller: inMemoryRadioGroupValue
+                ? (databasePathController..text = '')
+                : (databasePathController..text = widget.directory.path),
+            onFieldSubmitted: (value) async =>
+                await handleDatabasePathSubmitted(value),
+          ),
```
guisedlythe: Maybe add a validator that at least denies the empty path.

lib/utils/platform_database_factory.dart
```diff
@@ -0,0 +1,78 @@
...
+import 'package:flutter/foundation.dart';
+import 'package:path/path.dart' as path;
+import 'package:path_provider/path_provider.dart';
+import 'package:sembast/sembast.dart';
+import 'package:sembast/sembast_io.dart';
+import 'package:sembast/sembast_memory.dart';
+import 'package:sembast_web/sembast_web.dart';
+
+class PlatformDatabaseFactory extends DatabaseFactory {
+  String _databaseName = '';
```
guisedlythe: A bit dangerous that you need to set the database name, otherwise it will just use `.json`. Maybe use a default or init with null and check if it was set in `_finalPath()`?

guisedlythe (changes requested):
Lots of changes yet again... As I noted in some spots, please try to only modify and add the changes that are actually relevant for this PR.

I only very quickly went through it and it looked pretty good for the most part.
The most relevant thing I noticed is that the names resets to null when you do the setting reset.

lib/screens/settings/settings_screen.dart
JacobRoe: It actually is. Down below I use a `setState` call to force a rebuild. This is necessary because the different settings sections are dependent on many different services. An alternative would be to get all services and manually invoke the `invokeCallback` method but I think an empty `setState` is a more viable option in this case.

lib/screens/settings/database_list_section.dart
JacobRoe: In that case it's ok because it wipes the whole database and guides the user afterwards to the setup screens again. You probably mean the settings reset. That's in the `settings_screen`. (And yeah, you are right, I also noticed this issue when I showed you the reset.)

lib/screens/devices/device_list_screen.dart
JacobRoe: Good catch! 😅

lib/config/style_constants.dart
JacobRoe: I actually now follow the rules from the Dart style guide.

https://dart.dev/guides/language/effective-dart/style#dont-use-prefix-letters
https://dart.dev/guides/language/effective-dart/style#prefer-using-lowercamelcase-for-constant-names

lib/screens/activity_log_screen.dart
JacobRoe: Sorry. On windows the assets couldn't be loaded. I forgot to remove this change again and investigate.

JacobRoe:
> Lots of changes yet again... As I noted in some spots, please try to only modify and add the changes that are actually relevant for this PR.

I totally agree with you on this. I just like to allow myself the luxury to be driven by madness on academic projects. It always worked out great in the end.

Nonetheless, most issues you commented on should be resolved now. About the `DatabaseService` related classes I'm still not happy with but at least it should work as intended.

Thanks for your review!

guisedlythe (changes requested):
I just noticed that I get an error when running `flutter pub get` on Windows because of package version conflicts. Changing the `dart_code_metrics` version to `4.0.0` in `pubspec.yaml` fixed it for me.

I also found some missing comma warnings in the tests that I missed earlier 😅:
test\services\database_service_test.dart:
> Warning Prefer trailing comma. : 84:49 : prefer-trailing-comma https://github.com/dart-code-checker/dart-code-metrics/blob/master/doc/rules/prefer-trailing-comma.md
Warning Prefer trailing comma. : 100:53 : prefer-trailing-comma https://github.com/dart-code-checker/dart-code-metrics/blob/master/doc/rules/prefer-trailing-comma.md

test\utils\data_model_repository_test.dart:
> Warning Prefer trailing comma. : 16:54 : prefer-trailing-comma https://github.com/dart-code-checker/dart-code-metrics/blob/master/doc/rules/prefer-trailing-comma.md
Warning Prefer trailing comma. : 25:9 : prefer-trailing-comma https://github.com/dart-code-checker/dart-code-metrics/blob/master/doc/rules/prefer-trailing-comma.md
Warning Prefer trailing comma. : 40:9 : prefer-trailing-comma https://github.com/dart-code-checker/dart-code-metrics/blob/master/doc/rules/prefer-trailing-comma.md
Warning Prefer trailing comma. : 61:9 : prefer-trailing-comma https://github.com/dart-code-checker/dart-code-metrics/blob/master/doc/rules/prefer-trailing-comma.md

Other than that, I think it looks good!

Tan-jay (commented):
Code looks good! I really like the viewmodel :)
Just noticed following when testing:

* when reseting the database, the server cannot be started afterwards:  
Bildschirmfoto 2021-06-29 um 12 03 38

-> it needs a restart (or hot reload) in order to work again (i tested on my iPhone + mac desktop) 

* when starting the app on macOS desktop app for the first time, it is not asking for the database path (i deleted the app + p2p_task.json - maybe i forgot to delete something?)

* when reseting the settings the Activate server status is the same - should this also be set to off or on? (just thought as this is the only part in the synchronisation which is not set to something)

* when activating the server in settings, the IP Address under network is not shown, only when you click on the qr button (but i guess this was like that the whole time)

guisedlythe (approved):
Looking good. I will update the version for dart_metrics later in another PR.
Sorry for the late review and thank you for your patience. We can finally merge this!

PR #21, opened 2021-06-27
OPTM-58 Show error dialog when the server cannot be started from the settings screen

guisedlythe:
- Currently shows just uses the error.toString() as message


JacobRoe (commented):
Looks good! Just two minor comments.

(Just noticed it afterwards that it's a draft.)

lib/screens/error_popup_dialog.dart
```diff
@@ -0,0 +1,23 @@
+import 'package:flutter/material.dart';
+import 'package:flutter/widgets.dart';
+
+class SimpleErrorPopupDialog extends StatelessWidget {
```
JacobRoe: Maybe just me: By a (unspoken) convention a file name corresponds to the name of the main class inside the file. Although not required, I prefer it to be consistent.

lib/screens/settings/sync_list_section.dart
```diff
@@ -46,7 +47,17 @@ class SyncListSection extends StatelessWidget {
...
+                        ? peer
+                            .startServer()
+                            .onError((error, stackTrace) => showDialog(
+                                  context: context,
+                                  builder: (context) => SimpleErrorPopupDialog(
+                                    'Could not start server',
+                                    error.toString(),
+                                  ),
+                                ))
+                        : peer.stopServer(),
```
JacobRoe: The `onChanged` expression is quite large now. Maybe it makes sense to extract this into an own method inside the class. What do you think?

JacobRoe (approved):

PR #22, opened 2021-06-27
OPTM-78 Activity data

guisedlythe:


JacobRoe (changes requested):
1. I noticed that an overflow rendering exception happens on my device because the peer id takes more horizontal screen space than is available. It might won't happen that often anymore as soon as the name is used for that but it would still be possible (e.g. a name is chosen like "The peach tree from my neighbors garden stole my iPad device"). There are several options in Flutter available to prevent an overflow from happening. The easiest one would likely be to use the `overflow` option of the `(Rich)Text`-widget. Probably in combination with a widget like `Expanded`. (`Overflow.ellipsis` would replace parts of the string that overflows with three points instead.) It also overflows when the task or list name is too long. Maybe wrapping the timestamp to the next lines or explicitly keep the timestamp on the next line.

1. Furthermore, I created a task and instantly updated it but the order in the activity log seems wrong. It tells me that I updated the task before creating it. Both have the same date and time in the activity log, so my suspicion would be that the stored timestamp either only has precision int the seconds or something in the order method went wrong.

1. One feature request I might have been is to see the exact changes of an update activity log entry (maybe at least the title of a task or list). Currently, it offers me not much insight beside that a task or list was changed. But I don't know how feasible and/or expensive it would be to implement that. (Kinda' related to the next point, I noticed.)

1. Another questions: You mentioned that you can't show which element was deleted because it doesn't contain the value information in the CRDT-tree anymore. My first thought was it should be recoverable by the previous activities (e.g. finding the created event with the same id) but then I saw you create the activity entries on the fly based on the task-related CRDT, if I'm not mistaken. Therefore, the activity log loses any previously present information of that element as soon as it is deleted. I don't know how you think about that, but I wouldn't expect that behavior as a user of the app. On the other side, it would probably take too much time now to implement it differently.

1. In general, the `TaskListService` seems to do too much stuff currently. It contains the business logic for tasks and task list, CRDT-related methods and activity generation.

1. I also think the `ActivityEntry` is not ideal in its current state. Let's assume the app later adds several more features like customizable combi-filters, tags, sub-tasks or projects (which are also all shared with other devices) and you want changes to them appear in the activity log too. With your current approach you would need to add another field for their ids, and a corresponding getter for the check which class of activity it is. I assume we would use uuid's for every activity class, meaning a single field for an id suffice. Additionally, an enum for the different activity classes. Or even just subclassing. What do you think about it?

1. You could also extract the `ActivityType` enum and the classes `ActivityRecord` and `TaskListActivity` from the `task_list_service.dart` into their own files.

1. Another point, I found several minor naming issues, although it's probably just me. Throughout the code base, we typicially used the prefix `_build*` to denote methods that build sub widgets of the whole widget. In your case you began to use `_make*` instead. The former seems more "Flutteric" to me. Plus: I think almost all method names begin with a verb but your comparison method begins with a noun (e.g. `_activityEntryCompare` -> `_compareActivityEntry`, see also [Dart Style Guide](https://dart.dev/guides/language/effective-dart/design#prefer-an-imperative-verb-phrase-for-a-function-or-method-whose-main-purpose-is-a-side-effect)). Furthermore, you used a variable named `currentPeerId`. Should it be named `currentPeerID`? (I still prefer the former version, as does the [Dart Style Guide](https://dart.dev/guides/language/effective-dart/style#do-capitalize-acronyms-and-abbreviations-longer-than-two-letters-like-words), so I wouldn't mind if you keep the name the way it is. 😂) The suffix `Str` in `_getActivityDescriptionStr` can be omitted because the return type already provides this information for us.

1. Finally, you might be able to reduce some lines in the `TaskListService` class in the lines 236, 257, 275 and 286 for the creation of `TaskActivity` because the first three parameters always seem to be the same and only the ones after that change. But that's not really a must and I don't know if it really adds any value.

Otherwise, it looks really good. Btw: Although not strictly related to this PR, I love your test utility class `DeviceTaskList` a lot. Was very helpful to reduce all the previous boilerplate code in the tests.

guisedlythe:
I implemented most of your comments:

1. I used `Flexible` in the title and origin line so that long lines will wrap to multiple lines
2. That was a bug in the activity creation. I fixed it and also added that multiple activities for the same task can be generated.
3. Currently only the title of the task or task list is shown in the activity log. We could maybe add that the task list of a task is also displayed, but that would require an additional query of the task list database, which may be introduce even more work for the device. A cool future may be that you can click the activity and get sent straight to the task/list in question, we can maybe add that later.
4. Yeah, sadly that is not possible. We would need to store the activities locally, which we decided not to do for now. We could maybe add an information button at the top that explains this to the user because I agree that it can be a little counter-intuitive.
5. I agree, but I think refactoring the whole task list service is not part of this PR. We can do a PR especially for that purpose after this one (we wanted to adhere more to the code metrics anyway).
6. I completely removed the ActivityEntry class because it was an intermediate only used for the display of the activities in the task list screen anyway. The original purpose was probably to use it for storing the activity locally on the device, but we don't have that right now. The ActivityRecord class (which is based on subclassing) is used in its stead.
7. Done. Although I used one file for all activity record types for now. We only have two types and I didn't want to clutter the folder with too many tiny files. In future refactoring we should probably create a directory for all things related to the tasks and the crdt (doing that now would touch all files that include the task list service, which are quite a lot).
8. Done (I have no idea what I was thinking tbh, and let's just not talk about the `Id` :shushing_face:)
9. I refactored the activity generation into multiple methods, it should look better now.

I just remembered that I forgot the peer ID translation again...

guisedlythe:
I just noticed that the translation to peer names depends on peer identification, which should be possible after the security PR OPTM-50 (#19). I implemented it in the activity log so that it should translate the peer IDs to peer names once peer identification works and the peer IDs are 'synced' to the internal peer info IDs.

JacobRoe:
Saw your review request. Will try to look into it again this evening.

JacobRoe (dismissed):
LGTM

Works splendidly! I didn't thoroughly review the changes again because @firstdayofjune will also look into this PR, if I'm not mistaken. But from what I skimmed through of the code looked very good.

firstdayofjune (dismissed):
Did some refactorings and changed the UI slightly, since I now realized the Icons are proposed were not exactly ideal to support the understanding 😇 
Simulator Screen Shot - iPhone 12 Pro Max - 2021-07-09 at 12 39 52

firstdayofjune:
Looks good! I adapted the UI and at the same time refactored the widget. @PitchBlackCoffee may want to take a second look at the code now. Most of it should be fine, only thing to mention is that I added a getter for the description to the ActivityRecords. It returns a String only, in order to not prevent the standalone server.

Also, one thing I noticed is an update to the task list is logged as a create action. I couldn't figure out whether it is safe to add a isRecursiveUpdate parameter to the TaskListActivityRecord. If this can easily be done, could you fix this?

Minor remark: `activity_records.dart` feels rather like a model file, than a service. Maybe move it to the model folder.

PS. Regarding the ID vs Id, you know I can read this discussion? 😉 
Still, we discussed the naming as a matter of preference. Considering the reasoning in the dart style guide, now, I agree to use `Id` instead of `ID`, though. Let's vote on this again later in our meeting.

guisedlythe:
> Also, one thing I noticed is an update to the task list is logged as a create action. I couldn't figure out whether it is safe to add a isRecursiveUpdate parameter to the TaskListActivityRecord. If this can easily be done, could you fix this?

Your wish is my command. I also discovered a small bug because of this feature that I address in #30.

> Minor remark: activity_records.dart feels rather like a model file, than a service. Maybe move it to the model folder.

The `model` folder contains only files that can be serialized to json, which does (currently) not apply to the activity records. I would prefer to keep it together with the services, but add a folder for some of the services that require multiple files. We could do that in the future refactoring after we have done most of the features that we still want to implement.

> PS. Regarding the ID vs Id, you know I can read this discussion? wink
Still, we discussed the naming as a matter of preference. Considering the reasoning in the dart style guide, now, I agree to use Id instead of ID, though. Let's vote on this again later in our meeting.

Haha, sorry for that, I wouldn't have done it if you were not also assigned to this PR. I look forward to our repeated discussion though, then we should either change everything to one style or the other. We still use a mix of both styles in our code base because we didn't update our old code after our first discussion.

firstdayofjune (approved):

firstdayofjune:
Only thing is that already created tasks appear at the top of the list after updating a task list. But I guess this would be fixed with OPTM-93? So ready for merge.

guisedlythe:
> Only thing is that already created tasks appear at the top of the list after updating a task list. But I guess this would be fixed with OPTM-93?

Exactly.

PR #24, opened 2021-07-04
OPTM-49 Implemented OPTM-49 code review comments

guisedlythe:


lib/services/network_info_service.dart
```diff
@@ -20,29 +20,30 @@ class NetworkInfoService with LogMixin, ChangeCallbackProvider {
...
 
       return value;
     });
   }
 
   Future<String?> _detectSsid() async {
-    l.info('detecting ssid');
-    if (kIsWeb || !(Platform.isAndroid || Platform.isIOS)) return null;
+    logger.info('detecting ssid');
+    final isMobileDevice = Platform.isAndroid || Platform.isIOS;
```
JacobRoe: This line throws in the browser version because the browser doesn't have access to the Platform class (thus the network section in the settings screen just shows the text error). That's the reason why I put it after the `kIsWeb` check. A working alternative would be to make isMobileDevice a function that returns the result.

JacobRoe (commented):
Just skimmed through it really quick. Looks good overall but the "is mobile"-check in the `network_info_service.dart` file in the method `_detectSsid()` throws in the browser. See my comment for more infos.

firstdayofjune (approved):
Thanks for the changes.

PR #27, opened 2021-07-07
Fix code metrics warnings

JacobRoe:
OPTM-88, OPTM-89

The warnings are intentionally ignored.

The _long-method_ warnings are ignored in
- app_module.dart
- dependencies_provider.dart

because in both cases similar things are done in almost all lines. In the former file initializing the dependency graph for our app and in latter providing those dependencies into the widget tree. One could possibly reduce the lines either by grouping some parts or extracting every dependency into their own methods but I argue it would more likely reduce the readability in this case.

The _long-parameter-list_ warnings are ignored in
- platform_database_factory.dart
- yes_no_dialog.dart

because in the former file the class extends `DatabaseFactory` and overrides a method. Therefore, we can't change the method declaration. The latter, on the other hand, is a UI component and it's very common in Flutter to have a multitude of options for `Widget`s. (As a comparison see also the Flutter method `showGeneralDialog`, which even has a lot more options.)

The last case in the `SettingsScreen` widget is explained in an accompanying comment.

To my surprise those were the only warnings, thus I instantly fixed them all. (For that reason, this includes OPTM-89.)

### DIFF
--- a/lib/config/app_module.dart
+++ b/lib/config/app_module.dart
@@ -19,6 +19,7 @@ import 'package:sembast/sembast.dart';
 import 'package:shared_preferences/shared_preferences.dart';
 
 class AppModule {
+  // ignore: long-method
   Future<void> initialize(Injector injector) async {
     injector.dispose();
     injector = Injector();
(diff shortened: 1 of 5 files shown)


guisedlythe (approved):
LGTM.

Nice way of fixing code style warnings: just ignore them!
Jokes aside, I agree that all these warnings should just be ignored.
I also think that splitting the dependency and change notifier creation methods arbitrarily just to have less lines of code in one method does not improve our code quality. It is pretty stupid that the overridden method from an external package gives us a warning... I also don't see the reason for the long parameter list warning when you use named parameters that are not required.

PR #34, opened 2021-07-11
OPTM-75 Improved P2P

guisedlythe:
- A bit of refactoring in P2P related services (i.e. peer service, peer info service)
- Added (more) tests for the peer info service and peer service
- Can now have more than one peer location for a single peer (the functionality was there before, but very buggy and limited)
- Sync will try the locations consecutively. The peer location used for a successful sync will be moved to the front of the list
- Added button/tile in the device list screen to make adding locations to peer easier

Does not include:
- Detection if peer is already connected and using the existing connection
- Forwarding sync to other peers (broadcast / flooding of task list messages)


guisedlythe:
Discussed changes to the device list screen:
- Use collapsible cards to show peer infos and option to sync to specifc peer, delete it (only the information an uninformed user would need to know)
- Opening the collapsible shows more detailed information (i.e. peer locations, button to add new peer location to the peer info, ability to only try to sync with one specific peer location)

Some bugs:
- Adding a new peer resets the peer Status

guisedlythe:
I changed the device screen to use tiles. It now looks like this:
device_list

guisedlythe:
This PR now includes #33 because the list not refreshing on  some changes was very annoying

PR #35, opened 2021-07-11
OPTM-83 Error popup no sync

guisedlythe:
This PR uses #34

Adds an error popup when syncing to an active device or trying to connect to a created one fails.


lib/screens/devices/device_list_screen.dart
```diff
@@ -209,12 +321,67 @@ class _DeviceListScreenState extends State<DeviceListScreen> {
...
+            ? peerInfo
+            : peerInfo.copyWith(locations: [peerLocation]))
+        .then((success) {
+      if (!success) {
+        showDialog(
+          context: context,
+          builder: (context) => SimpleErrorPopupDialog(
+            'Sync Unsuccessful',
+            peerLocation == null
+                ? 'Could not sync with peer using "${peerInfo.name}"'
```
Tan-jay: could we change the message to 
```
? 'Could not sync with "${peerInfo.name}"'
: 'Could not sync with "${peerInfo.name}" using $location'
```
? 

and if there is an easy way to add the actual error like 'invalid location' or something? But don't waste your time on it, only when its quick

lib/screens/devices/device_list_screen.dart
```diff
@@ -209,12 +321,67 @@ class _DeviceListScreenState extends State<DeviceListScreen> {
...
+          : peerInfo.copyWith(locations: [peerLocation]),
+    )
+        .then((success) {
+      if (!success) {
+        showDialog(
+          context: context,
+          builder: (context) => SimpleErrorPopupDialog(
+            'Connection Unsuccessful',
+            peerLocation == null
+                ? 'Could connect to peer "${peerInfo.name}"'
```
Tan-jay: Could **not** connect to 😊
maybe also like this:
```
? 'Could not connect to "${peerInfo.name}"'
: 'Could not connect to "${peerInfo.name}" using ${peerLocation.uri}'
```

lib/screens/devices/device_list_screen.dart
```diff
@@ -158,28 +247,50 @@ class _DeviceListScreenState extends State<DeviceListScreen> {
...
+            ),
           ),
         IconSlideAction(
           caption: 'Delete',
           color: Colors.red.shade400,
           icon: Icons.delete,
-          onTap: () => viewModel.remove(peerInfo),
+          onTap: () => YesNoDialog.show(
+            context,
+            title: 'Delete Peer Location',
```
Tan-jay: maybe just "Delete Location" ?

lib/screens/devices/device_list_screen.dart
```diff
@@ -158,28 +247,50 @@ class _DeviceListScreenState extends State<DeviceListScreen> {
...
+            description:
+                'Do you really want to delete the peer location "${peerLocation.uri}" from the peer "${peerInfo.name}" with ID "${peerInfo.id}"?',
```
Tan-jay: unfortunately the text is too long for the showDialog
IMG_3434

this would work (considering a normal input) :
`'Do you really want to delete the location "${peerLocation.uri}" from "${peerInfo.name}"?'`

lib/screens/devices/device_list_screen.dart
```diff
@@ -117,31 +118,119 @@ class _DeviceListScreenState extends State<DeviceListScreen> {
...
+            textAlign: TextAlign.left,
+          ),
+        ),
+        Spacer(),
+        ..._buildSyncOrConnectButton(viewModel, peerInfo),
+        SizedBox(width: 5),
+        IconButton(
+          onPressed: () => YesNoDialog.show(
+            context,
+            title: 'Delete Peer',
```
Tan-jay: Better `Delete Device`

lib/screens/devices/device_list_screen.dart
```diff
@@ -158,28 +247,50 @@ class _DeviceListScreenState extends State<DeviceListScreen> {
...
-        'ID: ${peerInfo.id!}',
-      ),
+      child: _buildPeerLocationEntry(peerLocation),
+    );
+  }
+
+  Widget _buildAddPeerLocationTile(BuildContext context, PeerInfo peerInfo) {
+    return ListTile(
+      leading: Icon(Icons.add),
+      title: Text('Add Peer Location'),
```
Tan-jay: Just: `Add Location` or `Add Device Location` ?

lib/security/key_helper.dart
```diff
@@ -80,13 +80,20 @@ class KeyHelper with LogMixin {
...
+      try {
+        outputOffset += engine.processBlock(
+          input,
+          inputOffset,
+          chunkSize,
+          output,
+          outputOffset,
+        );
+      } on ArgumentError catch (e) {
+        // TODO: this is very implementation defined
```
Tan-jay: what does this mean? 
shouldn't it always throw an exception when this ArgumentError appears?

lib/screens/devices/device_list_screen.dart
```diff
@@ -158,28 +247,50 @@ class _DeviceListScreenState extends State<DeviceListScreen> {
...
+      onTap: () => _openDeviceForm(context, template: peerInfo),
```
Tan-jay, line 293: this opens the form with the title `Add Device`
IMG_3435

Maybe better to create a new form which just contains the IP Address and Location? Its a bit dangerous to change the Public Key by mistake

lib/screens/devices/device_list_screen.dart
```diff
@@ -117,31 +118,119 @@ class _DeviceListScreenState extends State<DeviceListScreen> {
...
+        ),
+        Spacer(),
+        ..._buildSyncOrConnectButton(viewModel, peerInfo),
+        SizedBox(width: 5),
+        IconButton(
+          onPressed: () => YesNoDialog.show(
+            context,
+            title: 'Delete Peer',
+            description:
+                'Do you really want to delete the peer "${peerInfo.name}" with ID "${peerInfo.id}"?',
```
Tan-jay: device instead of peer :)

lib/services/peer_info_service.dart
```diff
@@ -20,23 +20,62 @@ class PeerInfoService with ChangeCallbackProvider {
           .where((peerInfo) => peerInfo.id != null && peerInfo.name.isNotEmpty)
           .map((peerInfo) => MapEntry(peerInfo.id!, peerInfo.name)));
 
-  Future<void> upsert(PeerInfo peerInfo) async {
-    await _repository.upsert(peerInfo);
+  Future<void> upsert(
+    PeerInfo peerInfo, {
+    bool mergeWithExistent = true,
```
Tan-jay: this should always be true? mergeWithExistent can be removed and just checking for peerInfo.id != null :)

lib/services/peer_info_service.dart
```diff
@@ -20,23 +20,62 @@ class PeerInfoService with ChangeCallbackProvider {
...
+  }) async {
+    if (mergeWithExistent && peerInfo.id != null) {
+      await _repository.runTransaction((txn) async {
+        final existentPeerInfo = await _repository.get(peerInfo.id!, txn: txn);
+        if (existentPeerInfo != null) {
+          existentPeerInfo.locations
+              .forEach((location) => peerInfo.addPeerLocation(location));
+
+          // if existent peer info is active and the public key is the same, the new peer info should also be active
+          if (existentPeerInfo.status == Status.active &&
```
Tan-jay: This doesn't matter, the updated one should aways have the status of the existentPeerInfo.
A updates B's publicKey => B still has the correct publicKey of A, no introductionMessage (key exchange) needed

Tan-jay (commented):
- sorry for the insane huge images :D to lazy to resize now...

Really cool  :)
with the error pop up (OPTM-83) i had one issue: When you add an invalid ip address (spaces) you won't get a error message
IMG_3432 2
Bildschirmfoto 2021-07-19 um 23 18 28

for the device list (OPTM-114): mostly some things like wording changes, i would always go with device instead of peer (probably a normal user doesn't know what a peer is 😅) 

and OPTM-75: the "do you really want to delete" messages are too long and not everything is displayed and when there is an invalid location (with special characters) then the sync / updating the list doesn't work

lib/screens/devices/device_list_screen.dart
guisedlythe: I just marked the name, id, public key as readOnly and changed the title. I am too lazy to create a new form for this...

lib/screens/devices/device_list_screen.dart
guisedlythe: I removed the `maxLines: 4` for the `Text` in the YesNoDialog and added `scrollable: true` for the `AlertDialog`. It should now all be visible and in the worst case there should be a scroll bar.

guisedlythe:
Thank you for the detailed review.

As for the spaces in IP field there should now be the typical error when you try to sync. Additionally, IPs with spaces are now not allowed when creating new devices. Completely validating the IP/Domain is probably too difficult because domains can have very weird names.

lib/screens/devices/device_list_screen.dart
Tan-jay: 🙌 that's totally cool!

lib/screens/devices/device_list_screen.dart
Tan-jay: thats awesome! Could we just replace the peer again? 🙏
`'Do you really want to delete the location "${peerLocation.uri}" from the device "${peerInfo.name}" with ID "${peerInfo.id}"?'`

Tan-jay (dismissed):
thanks for the quick adaption! 
just one peer left, if you could replace it 🤓

and one more thing, when adding characters (invalid input) for the port you cannot submit it - also don't get an error message  (sorry this was the whole time like this and is not related to your PR, just tested it now) 
`Another exception was thrown: FormatException: Invalid radix-10 number (at character 1)`
if you could add a check there as well, would be awesome!!

but if there is no time left, let's just merge and i can have a look tomorrow as well for that :)

Tan-jay (approved):
perfect, thank you! 😊

PR #36, opened 2021-07-13
OPTM-116 test task related screens

firstdayofjune:
I added tests for task related screens, including:

- TaskListScreen
- TaskListsScreen
- TaskFormScreen

I also created a ~`utils` file in the test~ `widgets` file under `test/utils` with some helpers for testing. 

(Sidenote: I deleted the IncrementCounter example and the related test, for both is not used anymore and we have a number of widget tests now for future reference)


test/screens/task_list_screen_test.dart
```diff
@@ -0,0 +1,267 @@
...
+          await pumpAppAndSettle(tester, app);
+
+          final callToActionFinder =
+              find.text('Click the plus button below to add a ToDo.');
+          expect(callToActionFinder, findsOneWidget);
+        },
+      );
+
+      testWidgets(
+        'shows plus button to add a new list',
```
JacobRoe: The description doesn't conform to the tested behavior. It actually shows a button to add a new task instead of a list. I suppose, copy pasta 🍜 is the real evil 😈 here.

test/screens/task_lists_screen_test.dart
```diff
@@ -0,0 +1,177 @@
...
+      child: MaterialApp(
+        home: Scaffold(
+          body: TaskListsScreen(),
+        ),
+      ),
+    );
+  });
+
+  tearDownAll(() async {
+    await database.close();
```
JacobRoe: The `DeviceTaskList` class has a `close`-method. Thus, you could rather use that. (I think I saw the same thing before.) Would spare you a line.

test/utils/widgets.dart
```diff
@@ -0,0 +1,57 @@
...
+  await slideWidgetOpen(tester, slidable);
+}
+
+Future<void> pumpAppAndSettle(WidgetTester tester, Widget app) async {
+  await tester.pumpWidget(app);
+  await tester.pumpAndSettle();
+}
+
+Future<void> slideWidgetOpen(WidgetTester tester, Finder finder) async {
+  // If this test fails unexpectedly, the slideOffset is likely wrong.
```
JacobRoe: Could you elaborate on your comment?

JacobRoe (commented):
As always: I love your very structured and readable code. 👌

I added some minor comments but otherwise I think it's ready for merge.

test/screens/task_list_screen_test.dart
firstdayofjune: Nice catch, you are totally right!

test/screens/task_lists_screen_test.dart
firstdayofjune: Thanks for letting me know, I changed all my tests to use `DeviceTaskList.close` instead.

test/utils/widgets.dart
firstdayofjune: To slide the widget open, a certain threshold of sliding has to be surpassed.
I say "certain", because in the actual _flutter_slidable_ repo they use an offset of -50 to slide the widget open (https://github.com/letsar/flutter_slidable/blob/master/test/slidable_test.dart).
For some reason this did not work for us, so I tried different (much) higher offsets to see if this changes anything.
Eventually, I settled for -500.

I find this a little fragile and actually expect there is some constant hidden somewhere.
I just couldn't find it (yet) so I added the comment in case the slideWidgetOpen tests fail at anyone elses machine.

I updated the comment to make this more clear:
```
// If this test fails unexpectedly, the slideOffset is likely not "high"
// enough to swipe the widget open.
// Increase the offset e.g. to Offset(-1000, 0) and rerun the test.
``

firstdayofjune:
Thanks for your kind feedback!
It always takes me a couple of iterations and quite some time to refactor the code over and over again until I am satisfied with it (readability-wise).
So I really appreciate that the effort is well received 😊

I included your remarks and updated the comment on the slideWidgetOpen.
(I also merged main into my branch again.)

Let me know if there is anything else to change.

JacobRoe (approved):
LGTM 👍

